PR #73, opened 2015-01-07
Documentation updated/reformatting

moatazelmasry2:
Documentation addition. 
Added description to installation options. Formatted "Compilation options" to be sub paragraph of its previous paragraph

### DIFF
--- a/docs/install.markdown
+++ b/docs/install.markdown
@@ -25,6 +25,10 @@ not covered by the list above.
 
 ## Build/Install
 
+Currently Watchman can be installed either through: 
+1. Homebrew for OS X
+2. Source for other systems
+
 ### Installing on OS X via Homebrew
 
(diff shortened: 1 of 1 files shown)


docs/install.markdown
```diff
@@ -25,6 +25,10 @@ not covered by the list above.
 
 ## Build/Install
 
+Currently Watchman can be installed either through: 
+1. Homebrew for OS X
+2. Source for other systems
```
wez: This implies that we only support building via homebrew on OS X, which isn't the case.  Do we need to add this exact text?  Is there something you found unclear?

wez:
Sorry for taking a while to get around to this; I applied a slightly modified version of this via `hub am https://github.com/facebook/watchman/pull/73`.

Thanks!

PR #92, opened 2015-04-07
makefile: add missing destdir for install-exec-local target

seanfarley:
Seems to be a slight oversight that ${DESTDIR} was missing from this target.


wez:
Good catch; thanks!
applied via `hub am https://github.com/facebook/watchman/pull/92`

PR #94, opened 2015-04-08
Please do not put the state dir in /usr/var

okket:


okket:
ah, forget it, your code is correct, just override it with ```configure```. Sorry for the trouble.

PR #131, opened 2015-08-14
Specialized exception for socket timeouts

kwlzn:
Add a new specialized exception to pywatchman to help isolate routine socket timeouts from actual error conditions. This allows consumers to write non-blocking client code that properly handles error cases.

This should be a fully backwards compatible change and adds a test to prove that.

### DIFF
--- a/python/pywatchman/__init__.py
+++ b/python/pywatchman/__init__.py
@@ -40,6 +40,16 @@ class WatchmanError(Exception):
     pass
 
 
+class SocketTimeout(WatchmanError):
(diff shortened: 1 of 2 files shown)


wez:
`hub am`'d it and pushed; thanks!

kwlzn:
@wez thx! would it be possible to get a pywatchman release bump w/ this change?

wez:
Yep, I'll bump it a little later tonight

wez:
Pushed 1.2.0 to pypi

PR #136, opened 2015-08-18
Add `sync_timeout` option to `cmd_clock`

bmdhacks:
Problem

When doing large changes to the filesystem with git, the amount of
changes received from the next watchman query take a long time to
transfer.  This is despite the fact that we know of all these changes
since we were the ones performing them.

Solution

Add a 'sync_timeout' option to 'clock' which syncs the root and returns a clock
representing the current time.  This is effectively fast-forwarding all
the updates we just performed.


wez:
Thanks for contributing code!

I think I see what you're shooting for.  Can you elaborate on how you're using watchman currently?
The query command has an existing `sync_timeout` field that you can control and I wonder if that helps or how it doesn't quite help you.

bmdhacks:
Some bakground.  This is in the context of twitter's usage of watchman with git:

We've discovered a bug in OSX where big changesets confuse fsevents such that certain operations are not reported, but the flags are also not set indicating that something has been missed.  I can reproduce this by doing a git-pull over a years worth of the twitter repo, but I've been unable to separate it in a reproducible test-case that I can distribute.  I've confirmed with my own fsevents code that this is not a watchman bug.  It seems to be related to symlinks, and the fact that git creates symlinks before it creates the files they point to.

Furthermore, when I do a big git-pull, it creates a huge list of changes that have to round-trip from:
git-pull --> OS --> fsevents --> watchman --> git-status
This is stupid since git knows the files it just changed (and is more accurate since the above OSX symlink bug may arise).  

The solution is to use this supplied 'sync' command to watchman, where we do a watchman 'since', record the state, write out the git files and record those, then 'sync' with watchman to get the proper clock.  This is better than a 'since' because we don't want to transfer the ton of changes that we already know about.

Hope that clarifies what I'm aiming for.

wez:
OK, how about extending the `clock` command to specify an optional synchronization delay:

`["clock", {"sync_timeout": 100}]`

The default for sync_timeout can be 0, in which case the sync is skipped.

cmds/watch.c
```diff
@@ -27,25 +27,43 @@ bool w_cmd_realpath_root(json_t *args, char **errmsg)
...
+  unsigned int sync_timeout = 0;
+
+  if (json_array_size(args) == 3) {
+    json_t *options = json_array_get(args, 2);
+    if (options) {
+      json_t *st_json = json_object_get(options, "sync_timeout");
+      if (st_json) {
+        sync_timeout = (unsigned int)json_integer_value(st_json);
+      }
+    }
```
wez: you can use something like:

```
json_unpack(args, "[s, s, {s?:d}]", &ignored, &ignored, "sync_timeout", &sync_timeout);
```

to express this a bit more tersely.  Take a look at `cmd_version` in `cmds/info.c` for an example

cmds/watch.c
```diff
@@ -27,25 +27,43 @@ bool w_cmd_realpath_root(json_t *args, char **errmsg)
...
 
+  /* resolve the root */
   root = resolve_root_or_err(client, args, 1, false);
   if (!root) {
     return;
   }
 
+  if (sync_timeout) {
+    w_root_sync_to_now(root, sync_timeout);
+    root->ticks++;
```
wez: ticks will get increased automatically when the sync cookie is observed, so we shouldn't need this.  Also, if we wanted to do this here, we must hold the root lock when we increment it.

tests/integration/clock.php
```diff
@@ -13,4 +13,18 @@ function testClock() {
     $this->assertRegex('/^c:\d+:\d+:\d+:\d+$/', $clock['clock'],
                        "looks clocky");
   }
+
+  function testClockSync() {
```
wez: would you mind adding this as a new test in the python test suite instead?  We want to move away from the php based tests.

You could add `test/integration/test_clock.py` with just the python equivalent of this part in it.  Take a look at `tests/integration/test_since.py` for an example.

website/_docs/cmd.clock.markdown
```diff
@@ -8,8 +8,13 @@ permalink: docs/cmd/clock.html
 
 Returns the current clock value for a watched root.
```
wez: Insert:

```
*Since 3.8.*

*The [capability](/watchman/docs/capabilities.html) name associated with this
enhanced functionality is `clock-sync-timeout`.*
```

cmds/watch.c
```diff
@@ -27,25 +27,43 @@ bool w_cmd_realpath_root(json_t *args, char **errmsg)
   return true;
 }
```
wez: Insert:

```
W_CAP_REG("clock-sync-timeout")
```

website/_docs/cmd.clock.markdown
```diff
@@ -8,8 +8,13 @@ permalink: docs/cmd/clock.html
...
 
-Be careful how you interpret this value; it returns the instantaneous value of
-the clock, and may have changed by the time you intend to act upon it.
+If a sync_timeout option is supplied, the command blocks for the given
+number of microseconds waiting for all pending changes to sync.  The
```
wez: the timeout is measured in milliseconds.

The description of what this does is a little inaccurate.  It's probably a good idea to link to http://facebook.github.io/watchman/docs/cookies.html#how-cookies-work but summarize it here as:

sync_timeout specifies the number of milliseconds that you want to wait to observe a synchronization cookie.  The synchronization cookie is created at the start of your `clock` call and, once the cookie is observed, means that the clock value returned by this command is at least as current as the time of your `clock` call.

If no sync_timeout is specified, the returned clock value is the instantaneous value of the clock associated with the watched root, and may be almost immediately invalidated if there are any filesystem notifications that are yet to be processed.

cmds/watch.c
```diff
@@ -27,25 +27,43 @@ bool w_cmd_realpath_root(json_t *args, char **errmsg)
...
   }
 
+  /* resolve the root */
   root = resolve_root_or_err(client, args, 1, false);
   if (!root) {
     return;
   }
 
+  if (sync_timeout) {
+    w_root_sync_to_now(root, sync_timeout);
```
wez: part of me thinks that we should return an error response if this returns false.  Thoughts?

bgertzfield:
I'm super excited about this. The Buck project's daemon can really use this feature for its initial "establish a clock" feature.

bmdhacks:
Cool, I'll knock out the feedback issues and put an updated patch up today.

cmds/watch.c
bmdhacks: From my testing, w_root_sync_to_now pre-increments ticks, such that when the cookie event arrives, if it has many other changes inside it, then that tick will be wrapped up with those changes.  So if we modify 10,000 files and then sync, it's possible we'd get the clock that contains those 10,000 changes because clock ticks seem to be inclusive.

However I could be totally wrong about how this works, that's just how it seemed to me.

tests/integration/clock.php
bmdhacks: I'm going on vacation for three weeks in a few days, so this will have to wait.

bmdhacks:
Only the project owner can rebuild... can you click it?

bhamiltoncx:
@bmdhacks, mind fixing the failed builds?

wez:
I pushed your changes + some fixes as part of #160; thanks!

PR #138, opened 2015-08-22
allow consuming notifications while root lock is held

wez:
Summary: The goal is to minimize the risk that we encounter a
notification queue overflow.

This factors out our pending-notification state into a type
that represents a collection of pending notifications.  This collection
has its own locking and can be waited upon and signalled.

The concept is that a notification thread will maintain a local
collection and accumulate pending paths until it reaches our
batch size or has consumed all current available notifications.

At that point, it locks the root->pending collection and moves its local
collection into root->pending, and then signals and unlocks
root->pending.  The notification thread can then continue to consume and
accumulate more notifications into its local collection.

The io thread also maintains a local collection.  It waits for
root->pending to become signalled, locks it, and moves root->pending
into its local collection and unlocks root->pending.  The io thread
will then process its local collection before returning to wait on
root->pending.

Refs: https://github.com/ember-cli/stress-app/issues/1

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,7 @@ watchman_SOURCES = \
 	hash.c       \
 	ht.c         \
 	ioprio.c        \
+	pending.c       \
 	stream.c        \
 	stream_stdout.c \
 	stream_unix.c   \
--- a/pending.c
+++ b/pending.c
@@ -1,0 +1,199 @@
+/* Copyright 2012-present Facebook, Inc.
+ * Licensed under the Apache License, Version 2.0 */
+
+#include "watchman.h"
+
+/* Free a pending_fs node */
+void w_pending_fs_free(struct watchman_pending_fs *p) {
+  w_string_delref(p->path);
+  free(p);
+}
+
+/* initialize a pending_coll */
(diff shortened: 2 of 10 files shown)


stefanpenner:
Oh man, I really appreciate you digging in:)

stefanpenner:
In the interim, is there a way for us to squelch the output in this scenario?

wez:
Pushed as https://github.com/facebook/watchman/commit/c650e85d27458df0e70aeab57c2890b8dae286ee

stefanpenner:
@wez I still seem to have the warnings, although now just when re-connecting to an existing watch-project.

```
watchman watch-del-all
ember s
// ok
<ctrl-c>
ember s

Recrawled this watch 6 times, most recently because:
/Users/stefanpenner/tmp/slow-ember-cli-project: kFSEventStreamEventFlagUserDropped
To resolve, please review the information on
https://facebook.github.io/watchman/docs/troubleshooting.html#recrawl
To clear this warning, run:
`watchman watch-del /Users/stefanpenner/tmp/slow-ember-cli-project ; watchman watch-project /Users/stefanpenner/tmp/slow-ember-cli-project`
... etc
```

PR #145, opened 2015-09-11
Add the "foreign" flag to Makefile.am

nodakai:
The structure of the watchman package does not conform to GNU's packaging standard;  namely it lacks files such as `NEWS`, `README` (not `README.md`) and Autoconf is not happy about it.

- http://stackoverflow.com/questions/15013672/use-autotools-with-readme-md

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,5 @@
+AUTOMAKE_OPTIONS = foreign
+
 bin_PROGRAMS = watchman
 # ensure that we have a reasonable default for the python install
 DESTDIR ?= /


wez:
Sorry for not getting back to you sooner.  We use `autogen.sh` to generate the makefile and pass the foreign flag in there.  ISTR that we had to pass the flag rather than include it in the makefile to avoid issues with older automake on some versions of Solaris.

I can look into whether this is still the case, but we recommend using `autogen.sh` in any case.

nodakai:
Understood you had compatibility issues in your mind.  Let me see how I can
use autogen.sh in Debian's debhelper workflow.

2015-09-18 5:18 GMT+08:00 Wez Furlong <notifications@github.com>:

wez:
I'm closing this out due to inactivity and because it should be solvable through other means.

PR #149, opened 2015-09-18
Set zip_safe = True in setup.py to avoid warning when building

bhamiltoncx:
When building pywatchman with python setup.py install, we get a warning:

```
zip_safe flag not set; analyzing archive contents...
```

I confirmed watchman is safe to run from inside a zip (Buck has to do this in its PEX archive), so this diff sets the flag to stifle the warning.


wez:
Pushed as a510de0b190ba182ad56b39f68572095cbccd2e4, thanks!

PR #150, opened 2015-09-21
race: two clients can race and start separate watchers for the same dir

wez:
Seems to happen rarely, but I have seen the logs that prove it.
The remediation is pretty simple; we just need to check to see if
someone beat us to registering the root that we're creating and
return that one instead.

### DIFF
--- a/root.c
+++ b/root.c
@@ -2208,7 +2208,7 @@ bool w_is_path_absolute(const char *path) {
 static w_root_t *root_resolve(const char *filename, bool auto_watch,
     bool *created, char **errmsg)
 {
-  struct watchman_root *root = NULL;
+  struct watchman_root *root = NULL, *existing = NULL;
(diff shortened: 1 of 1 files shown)


root.c
```diff
@@ -2311,8 +2311,18 @@ static w_root_t *root_resolve(const char *filename, bool auto_watch,
   *created = true;
```
bhamiltoncx: I think it'd be clearer to move this down inside the `pthread_mutex_lock(&root_lock)` and setting it only in the `else` block. (It's not like a pointer dereference under the lock is going to cost much at all compared to the `w_ht_get()` and `w_ht_set()` calls..)

bhamiltoncx:
Looks pretty good to me! I just had the one nit-pick.

root.c
wez: good call; updated!

PR #151, opened 2015-09-22
Improve PCRE detection in configure.ac

bhamiltoncx:
Previously, calling `configure --with-pcre` would result in a Watchman build silently succeeding but failing to include PCRE if `pcre-config` printed `CFLAGS` in a non-standard location.

This was because we failed to set `CPPFLAGS` before calling `AC_CHECK_HEADERS` (that macro does not use `CFLAGS` for some unknown reason).

I fixed this as well as improved PCRE detection by:

1. Failing fast if `pcre-config` could not be found if `--with-pcre` is specified
2. Printing the version of PCRE if found
3. Setting `CPPFLAGS` to ensure `AC_CHECK_HEADERS` works correctly with headers in a nonstandard location
4. Failing fast if `pcre-config` emits flags which don't work

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,7 +95,6 @@ AC_ARG_WITH(ruby, [
 ])
 AM_CONDITIONAL(HAVE_RUBY, [test "x$want_ruby" = "xyes"])
 
-pcre_config="yes"
 AC_ARG_WITH(pcre, [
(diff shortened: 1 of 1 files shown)


wez:
pushed as 99dde8ee3f13233be097c036147748b2d7f8bfa7, thanks!

PR #157, opened 2015-10-02
fsevents coalesces directory renames to top level

wez:
Summary: OS X commonly saves "documents" as a directory containing a
collection of files.  When updating one of these documents, the pattern
is to build out a temporary directory containing the target set of files
and then to rename this directory in place of the old directory.

When fsevents reports these events, there is a good chance that the
events will get coalesced and reported as a rename event on the document
dir followed by change events on the temporary dir structure.

We have two problems in tracking and dealing with this scenario:

1. We didn't consider that the rename event on the document dir should
   be recursive, so we'd only add_pending and stat just the dir

2. Once making it recursive, we had an optimization in the `crawler`
   function that would only directly stat the contents if the file
   had transitioned between existing or not existing.
   This optimization is valid only if the watcher has
   per-file-notifications that are reliably reported for every file
   in a watched dir.

This diff adds a flag to the watcher layer to identify this
coalesce-on-rename behavior and fixes up the crawler optimization so
that we will add_pending if the watcher has the coalescing behavior.

Test Plan: `make integration`, but also tested Blair's repro scenario
and cross-checked the watchman results via:

```
watchman find . path/to/document/file
/usr/bin/stat -r path/to/document/file
```

### DIFF
--- a/root.c
+++ b/root.c
@@ -988,12 +988,15 @@ static void stat_path(w_root_t *root,
           // called for the root itself)
           w_string_equal(full_path, root->query_cookie_dir)) {
 
-        if (!watcher_ops->has_per_file_notifications) {
+        if (!watcher_ops->flags & WATCHER_HAS_PER_FILE_NOTIFICATIONS) {
           /* we always need to crawl, but may not need to be fully recursive */
           crawler(root, coll, full_path, now, recursive);
         } else {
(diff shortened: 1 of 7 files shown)


wez:
Pushed as 11b1f4abebebf4893e294c85314e27f80d9aa71d

PR #169, opened 2015-10-16
Employ very strict O_NOFOLLOW semantics when we opendir

wez:
Summary: There are a couple of scenarios where a dir can be replaced
with a symlink that can skirt around our checks for symlinks.

We want to avoid traversing symlinks as this can be a costly operation
that takes us across filesystem boundaries.

This diff adds a stricter open operation that exclusively applies
O_NOFOLLOW to each component of the path that we want to open.
If any component of this fails, then the entire open operation fails.

We use this as part of our opendir_nofollow function.

We'll catch the failure to open in the various watcher
`root_start_watch_dir` functions, which then call `handle_open_errno`
and will recursively prune out the nodes from the tree.

Test Plan: `make integration` and run on my devserver.  Relatively lame.

### DIFF
--- a/opendir.c
+++ b/opendir.c
@@ -47,23 +47,113 @@ struct watchman_dir_handle {
   struct watchman_dir_ent ent;
 };
 
+/* Opens a directory, strictly prohibiting opening any symlinks
+ * in any component of the path on systems that make this
+ * reasonable and easy to do.
+ * We do this by opening / and then walking down the tree one
+ * directory component at a time.
+ * We prefer to do this rather than using realpath() and comparing
+ * names for two reasons:
+ * 1. We should be slightly less prone to TOCTOU issues if an element
+ *    of the path is changed to a symlink in the middle of our examination
+ * 2. We can terminate as soon as we discover a symlink this way, whereas
+ *    realpath() would walk all the way down before it could come back to
+ *    us and allow us to detect the issue.
+ */
+static int open_dir_as_fd_no_symlinks(const char *path, int flags) {
+#ifdef HAVE_OPENAT
+  char *pathcopy;
+  int fd = -1;
+  int err;
+  char *elem, *next;
+
+  if (path[0] != '/') {
+    errno = EINVAL;
+    return -1;
+  }
+
+  if (strlen(path) == 1) {
+    // They want to open "/"
+    return open(path, O_NOFOLLOW | flags);
+  }
+
+  pathcopy = strdup(path);
+  if (!pathcopy) {
+    return -1;
+  }
+
+  // "/" really shouldn't be a symlink, so we shouldn't need to use NOFOLLOW
+  // here, but we do so anyway for consistency with how we open the
+  // intermediate descriptors in the loop below.
+  fd = open("/", O_NOFOLLOW | O_CLOEXEC);
+  if (fd == -1) {
+    err = errno;
+    free(pathcopy);
+    errno = err;
+    return -1;
+  }
+
+  // "/foo/bar" means that elem -> "foo/bar"
+  elem = pathcopy + 1;
+  while (elem) {
+    int next_fd;
+
+    // if elem -> "foo/bar" we want:
+    // elem -> "foo", next -> "bar"
+    // if there is no slash, say if elem -> "bar":
+    // elem -> "bar", next -> NULL
+    next = strchr(elem, '/');
+    if (next) {
+      *next = 0;
+      next++;
+    }
+
+    // Ensure that we're CLOEXEC on all intermediate components of
+    // path, but use the flags the caller passed to us when we reach
+    // the leaf of the tree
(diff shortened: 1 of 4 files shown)


opendir.c
```diff
@@ -47,23 +47,91 @@ struct watchman_dir_handle {
...
+    errno = EINVAL;
+    return -1;
+  }
+
+  pathcopy = strdup(path);
+  if (!pathcopy) {
+    return -1;
+  }
+
+  fd = open("/", O_NOFOLLOW | O_CLOEXEC);
```
bhamiltoncx: If "/" is a symlink, we're in real trouble. Why bother with `O_NOFOLLOW` there?

opendir.c
```diff
@@ -47,23 +47,91 @@ struct watchman_dir_handle {
...
   return win_opendir(path, 1 /* no follow */);
 #else
-  int fd = open(path, O_NOFOLLOW | O_CLOEXEC);
+  int fd = open_dir_as_fd_no_symlinks(path, O_NOFOLLOW | O_CLOEXEC);
   if (fd == -1) {
     return NULL;
   }
-#if defined(__APPLE__)
+# if !defined(HAVE_FDOPENDIR) || defined(__APPLE__)
+  /* fdopendir doesn't work on OS X */
```
bhamiltoncx: It seems to work as of OS X 10.11:

```
% cat fdopendir.c
#include <dirent.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int main(int argc, char** argv) {
  int fd = open("/", 0);
  if (fd == -1) {
    perror("open");
    return 1;
  }
  DIR *dir = fdopendir(fd);
  if (dir == NULL) {
    perror("fdopendir");
    return 1;
  }
  struct dirent *dp;
  while ((dp = readdir(dir)) != NULL) {
    printf("%s\n", dp->d_name);
  }
  closedir(dir);
  close(fd);
  return 0;
}

% clang -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -o fdopendir fdopendir.c
% ./fdopendir
.
..
(snip a bunch of stuff)
```

opendir.c
```diff
@@ -47,23 +47,91 @@ struct watchman_dir_handle {
   struct watchman_dir_ent ent;
 };
 
+/* Opens a directory, strictly prohibiting opening any symlinks
+ * in any component of the path on systems that make this
+ * reasonable and easy to do.
+ * We do this by opening / and then walking down the tree one
```
bhamiltoncx: Could we just call `realpath()` and see if the result changes?

opendir.c
wez: In earlier versions, we had a configure check for it, it would link, but it would always error out :-/
However, we now use the bulk stat stuff since 10.10, so we avoid this issue in most systems on OS X

opendir.c
wez: We could, but I think this way is slightly less prone to TOCTOU and is possibly more efficient in case any element of the path is a symlink; we'll bail immediately, whereas the realpath approach would resolve all the way down before we got to compare the names.

opendir.c
wez: Yeah, it should be unpossible.  I don't think it hurts to add this check here and helps underscore the intent

opendir.c
bhamiltoncx: Can we add a configure check for the error, then? If it works in 10.11 we should use it.

opendir.c
bhamiltoncx: OK. Maybe worth a comment as to why we don't do that?

opendir.c
bhamiltoncx: Maybe worth a comment then.

opendir.c
wez: We use `getattrlistbulk` on 10.10 and up; I think it's wasted cycles to beef up that configure check

opendir.c
wez: Will do

opendir.c
bhamiltoncx: OK. More comments then please. :)

sid0:
Seems like this will add an O(log N) component to crawls -- have you checked its impact?

wez:
re: perf, yeah, it doesn't materially change the full crawl wall time, sharing a paste with you OOB from this PR

wez:
FWIW, I got this to prune out symlinks rebasing back and forth between an old branch and master:

```
2015-10-16T15:36:01,164: [io /REPO] opendir(/REPO/third-party-buck/build/THING/lib/python) -> Too many levels of symbolic links. Marking this portion of the tree deleted
```

bhamiltoncx:
Looks OK to me.

wez:
Pushed as f7addabede29270bf25b66c2114c26a00ef0f902

PR #170, opened 2015-10-17
Add watchman-make and watchman-wait

wez:
These tools should help with the most common casual usecases for
watchman.


wez:
Pushed as 5d3d60c

PR #171, opened 2015-10-20
Add periodic socket liveness check

wez:
Summary: since we're using unix domain sockets, it is possible for a
couple of funky situations to happen:

* Someone can remove the socket file.  Perhaps they are being
  overzealous in fixing some issue, or perhaps the socket gets
  nuked by something like tmpwatch

* It's conceivable that there could be a race condition with multiple
  clients trying to start the service at the same time.

The latter is a bug that we should also try to run down and resolve,
but it is worthwhile to detect and auto remediate if we detect that
the unix socket is broken.

This diff will periodically (every 10 seconds, or each time `poll(2)`
errors out), schedule a thread that will connect and sanity check
that `watchman get-pid` returns our own pid.  If we have any failure
to connect or a mismatch in the result then we'll self terminate.

While working on this I realized that we broke W_LOG_FATAL at some
point; we were just skipping FATAL level messages because the value
was outside the interesting range(!).   That might explain some
of the weird CPU spins we've seen in a couple of weird cases, where
we should have terminated but are still trying hard.

This fixes the value so that it is below the "off" threshold.

Test Plan:

Run this in one window:

`./watchman --logfile=/dev/stdout --foreground -U /tmp/foow`

Then run the same thing in another window.

The first one will output:

```
./watchman --logfile=/dev/stdout --foreground -U /tmp/foow
2015-10-20T13:16:03,838: [listener] Watchman 4.0.0 93e07ce578178c781981960cd7ff0047626f4e19 starting up on wez-mbp
2015-10-20T13:16:03,838: [listener] Using watcher mechanism fsevents
2015-10-20T13:16:03,838: [listener] file limit is 2560 kern.maxfilesperproc=10240
2015-10-20T13:16:03,838: [listener] raised file limit to 10240
2015-10-20T13:16:03,838: [listener] launchd checkin failed: No such process
2015-10-20T13:16:43,853: [sockcheck] remote pid from get-pid (51812) doesn't match my pid (51718)
2015-10-20T13:16:43,854: [sockcheck] Fatal error detected at:
2015-10-20T13:16:43,854: [sockcheck] 0   watchman                            0x00000001045796d0 w_log + 672
2015-10-20T13:16:43,854: [sockcheck] 1   watchman                            0x00000001045790b0 check_my_sock + 432
2015-10-20T13:16:43,854: [sockcheck] 2   libsystem_pthread.dylib             0x00007fff909259b1 _pthread_body + 131
2015-10-20T13:16:43,854: [sockcheck] 3   libsystem_pthread.dylib             0x00007fff9092592e _pthread_body + 0
2015-10-20T13:16:43,854: [sockcheck] 4   libsystem_pthread.dylib             0x00007fff90923385 thread_start + 13
[1]    51718 abort      ./watchman --logfile=/dev/stdout --foreground -U /tmp/foow
```

I also tried removing `/tmp/foow` with a similar result.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,6 +16,7 @@ watchman_SOURCES = \
 	spawn.c       \
 	opt.c        \
 	cfg.c        \
+	checksock.c  \
 	fstype.c     \
 	log.c        \
 	json.c       \
--- a/checksock.c
+++ b/checksock.c
@@ -1,0 +1,79 @@
+// Copyright 2004-present Facebook. All Rights Reserved
+
+#include "watchman.h"
+
+/* Periodically connect to our endpoint and verify that we're talking
+ * to ourselves.  This is normally a sign of madness, but if we don't
+ * get an answer, or get a reply from someone else, we know things
+ * are bad; someone removed our socket file or there was some kind of
+ * race condition that resulted in multiple instances starting up.
+ */
+
+static void *check_my_sock(void *unused) {
+  json_t *cmd = json_pack("[s]", "get-pid");
+  json_t *result;
+  w_stm_t client = NULL;
+  w_jbuffer_t buf;
+  json_error_t jerr;
+  json_int_t remote_pid = 0;
+  pid_t my_pid = getpid();
+
+  unused_parameter(unused);
+  w_set_thread_name("sockcheck");
+
+  client = w_stm_connect(get_sock_name(), 6000);
+  if (!client) {
+    w_log(W_LOG_FATAL, "Failed to connect to myself for get-pid check: %s\n",
+        strerror(errno));
+    /* NOTREACHED */
+  }
+
+  w_stm_set_nonblock(client, false);
+
+  w_json_buffer_init(&buf);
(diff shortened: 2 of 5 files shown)


bhamiltoncx:
If the intention is to fix the race condition, I'd feel a lot more comfortable doing that.

I hate waking up unnecessarily and talking to the filesystem, since it's going to spin up hard disks to walk the filesystem metadata even if nothing has changed.

wez:
Nod.  While looking at this, I noticed that we can detect the removal by poll erroring out with ENOENT, so I think that I'll keep this check (I still think it is good to catch folks nuking things), but increase that timeout to something like 24 hours so that we avoid too much in the way of wakeups

wez:
Turned out I was wrong about the ENOENT thing.  I raised the poll timeout to 60 seconds; there isn't any other way to cheaply check whether it has been removed, so this is a reasonable compromize.

Pushed as 97d997940640e0c95e368e8f6d3f349a2375984d

PR #173, opened 2015-10-21
Fix st_size for symlinks when using bulkstat

wez:


wez:
Pushed as dbc810f777b4fd84c9c1fe654591aa3f4fb12ef6

PR #174, opened 2015-10-21
pywatchman: add immutable version of data results

wez:
This adds an alternative object representation for decoded bser
values.   The new representation is immutable and can have its
properties accessed via either the mapping protocol or as attributes.

The goal is to eliminate an intermediate representation in mercurial
and just return the bser values directly in place of stat-like
objects that are manufactured by the hgwatchman extension.

This can save several seconds of runtime in hgwatchman if we adjust it
to process fresh_instance result sets with very large trees.

`bser.loads` has been augmented with an optional `mutable` flag that
defaults to `True`.   When `True` we return the same dicts and lists
of dicts that we've always done.  When `False` we return our immutable
object representation.

pybser has been updated to respect this flag also.

Test Plan: `make integration`.  I also built a debug cpython with ASAN
to verify the reference counting.

### DIFF
--- a/python/pywatchman/pybser.py
+++ b/python/pywatchman/pybser.py
@@ -213,45 +213,106 @@ def _bunser_string(buf, pos):
     return (str_val, pos + str_len)
 
 
-def _bunser_array(buf, pos):
+def _bunser_array(buf, pos, mutable=True):
     arr_len, pos = _bunser_int(buf, pos + 1)
     arr = []
     for i in range(arr_len):
-        arr_item, pos = _bser_loads_recursive(buf, pos)
+        arr_item, pos = _bser_loads_recursive(buf, pos, mutable)
         arr.append(arr_item)
+
+    if not mutable:
+      arr = tuple(arr)
+
     return arr, pos
 
 
-def _bunser_object(buf, pos):
+# This is a quack-alike with the bserObjectType in bser.c
+# It provides by getattr accessors and getitem for both index
+# and name.
+class _BunserDict(object):
+    __slots__ = ('_keys', '_values')
+
+    def __init__(self, keys, values):
+        self._keys = keys
+        self._values = values
+
+    def __getattr__(self, name):
+        try:
+            return self._values[self._keys.index(name)]
+        except ValueError as ex:
+            raise AttributeError('_BunserDict has no attribute %s' % name)
+
+    def __getitem__(self, key):
+        if isinstance(key, (int, long)):
+            return self._values[key]
+        elif key.startswith('st_'):
+            # hack^Wfeature to allow mercurial to use "st_size" to
+            # reference "size"
+            key = key[3:]
+        try:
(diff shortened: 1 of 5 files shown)


python/pywatchman/pybser.py
```diff
@@ -213,45 +213,108 @@ def _bunser_string(buf, pos):
...
+
+    if not mutable:
+      arr = tuple(arr)
+
     return arr, pos
 
 
-def _bunser_object(buf, pos):
+# This is a quack-alike with the bserObjectType in bser.c
+# It offers the mapping protocol via the dict base class and
```
wez: will update the comment to reflect the current implementation

python/tests/tests.py
```diff
@@ -98,9 +111,13 @@ def test_template(self):
         exp = [
             {"name": "fred", "age": 20},
             {"name": "pete", "age": 30},
-            {"age": 25}
+            {"name": None, "age": 25}
```
wez: this is the only behavioral change; previously we would not set `name` if we were SKIPing a value.  Nothing generated this in practice and it makes things more regular if we return None instead

python/pywatchman/pybser.py
```diff
@@ -213,45 +213,106 @@ def _bunser_string(buf, pos):
...
+      arr = tuple(arr)
+
     return arr, pos
 
 
-def _bunser_object(buf, pos):
+# This is a quack-alike with the bserObjectType in bser.c
+# It provides by getattr accessors and getitem for both index
+# and name.
+class _BunserDict(object):
```
bhamiltoncx: How about `collections.Mapping` instead? It's immutable and will provide you with more useful dict-like methods.

python/setup.py
```diff
@@ -22,7 +22,11 @@
     license = 'Apache License (2.0)',
     packages = ['pywatchman'],
     ext_modules = [
-        Extension('pywatchman.bser', sources = ['pywatchman/bser.c'])
+        Extension('pywatchman.bser', sources = ['pywatchman/bser.c']
+        #    ,
+        #    extra_compile_args=['-O0', '-g'],
```
bhamiltoncx: Did you mean to check this in?

python/pywatchman/pybser.py
wez: My first pass at this just used dict and added the getattr out of laziness, but then I realized that I should make it more like the thing we're exporting from the native C extension to minimize the chance they pybser behaves differently from the native extension.  I'm still pretty new to the python internals so I don't know exactly how close or how far from that goal we are here :-)

python/setup.py
wez: yes and no... it was handy while debugging.  I'll tidy this up and/or remove it before landing

wez:
Pushed as 938cbcc60aa7801ecb5d7faa6d46f674e6b15c25

PR #177, opened 2015-10-23
Only build Python bindings if Python.h was found

dschmidt:
Hey,

I'm really not familiar with autotools, so bear with me, if this is not the correct approach.
Just let me know and I will fix...

BR,
Dominik

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,8 +84,19 @@ AC_ARG_WITH(python, [
 ])
 if test "x$want_python" = "xyes" ; then
   AM_PATH_PYTHON(,,[:])
+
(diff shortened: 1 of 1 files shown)


wez:
Can you share some more details about the problem that you're aiming to fix here?  Missing python-devel packages?

dschmidt:
This was declined on IRC, because succefully checking for Python.h might need environment setup by setup.py.
At this point I don't want to waste anymore time on this. Just squashed commits and leaving it in a clean state if anyone wants to carry on or give it a go him/herself. Have fun!

wez:
Thanks for taking a crack at this!

PR #183, opened 2015-10-29
Fix range check in _int_size

### LINKED ISSUE #180: Off by one error in _int_size
I noticed this when I was in the area from #179, but the \_int_size calculations have an off by one error: The smallest representable twos-complement integer in n bytes is -2^(n - 1), but this function treats it as if it were  - 2^(n - 1) - 1 (i.e. the negation of the largest reprsentable number).

This means that some numbers require one more byte to represent than is necessary (not a major deal) but also that it errors erroneously about -2^63.

bhamiltoncx:
Fixes #180.

Test Plan: `python runtests.py`

I could probably write better tests for this, but this fixes the important issue (cannot encode `-0x8000000000000000L`).

### DIFF
--- a/python/pywatchman/pybser.py
+++ b/python/pywatchman/pybser.py
@@ -52,14 +52,13 @@ EMPTY_HEADER = "\x00\x01\x05\x00\x00\x00\x00"
 
 def _int_size(x):
     """Return the smallest size int that can store the value"""
-    abs_x = abs(x)
-    if abs_x <= 0x7F:
+    if -0x80 <= x <= 0x7F:
         return 1
-    elif abs_x <= 0x7FFF:
+    elif -0x8000 <= x <= 0x7FFF:
         return 2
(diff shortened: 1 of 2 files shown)


wez:
I think you've found a bug in the native module on windows (yay!)
https://ci.appveyor.com/project/wez/watchman/build/b154#L307
I suspect something to do with LLP64

bhamiltoncx:
Oy.
2015年10月31日(土) 7:39 Wez Furlong <notifications@github.com>:

wez:
74aff4684a0e5151e90e3ef27ef442a31e062219 addressed the native module issue.  Pushed this as 16ca76add96de462ac08b8c4019c6922f527c236

Thanks!

bhamiltoncx:
Cheers!
2015年10月31日(土) 10:06 Wez Furlong <notifications@github.com>:

PR #187, opened 2015-11-03
Add state-enter, state-leave commands for advanced settling

wez:
Full details on how to use it are in the docs commit.

The goal is to facilitate https://www.mercurial-scm.org/wiki/WorkingCopySyncPlan by allowing mercurial to tell watchman when it is about to update the working copy.

psuedo-code in the mercurial update path might go something like this:

```
client.query("state-enter", "/path/to/repo", {
   "name": "hg.update",
   "metadata": {
       "current_rev": "<hash>",
       "target_rev": "<hash>",
    }
})

perform_wc_update()
# get list of modified files
modified_files = do_hg_status()

client.query("state-leave", "/path/to/repo", {
   "name": "hg.update",
   "metadata": {
       "current_rev": "<hash>",
       "target_rev": "<hash>",
       # tell subscriber which files were modified since target_rev
       "files": modified_files
    }
})
```


stefanpenner:
Ah, this looks like something I can issue during before/after a build. To hint to watchman, preventing noise from  during a build from causing grief?

wez:
@stefanpenner potentially, yeah.  On the producer side, you can bracket you build with `state-enter ember.build` and `state-leave ember.build`.

If you have subscribers and want them to be build aware, you can have their subscriptions specify `defer: "ember.build"` or `drop: "ember.build"` in their subscription definition to defer or drop notifications while the tree is in that state.

The gain is that you can potentially reduce the amount of effort exerted by your subscriber; deferred notifications will be dispatched after the `state-leave ember.build` is issued and will likely coalesce the results.  Dropped notifications won't be sent at all while the state is asserted.

We don't (yet) have a way to express this same concept if you're polling via the `query` command.

stefanpenner:
@wez can't wait to find some time to give it a try :)

wez:
I pushed this to master, for better or worse!

Issue #172, opened 2015-10-20
Dangling symlinks missed in OSX


bmdhacks:
This is an OSX bug, but I figured I'd fill you guys in because it's causing Twitter all kinds of headaches.  Here's a repro:

```
#!/usr/bin/env bash

set -x

watchman watch .
watchman since . n:foo
ls -lh path0sym
mkdir path2 path3 path3/subp3
paths="path0 path2/file2 path3/file3 path3/subp3/file3"
(
  for p in $paths
    do
        echo "hello $p" >$p || exit 1
        ln -s "hello $p" ${p}sym || exit 1
    done
)
ls -lh path0sym
```

   watchman since . n:foo

Note that the second 'since' query doesn't return any of the *sym files that were created.  Change `"hello $p"` to `"$(basename $p)"` and it will work.

wez:
Thanks for reporting.  I beefed up your script a bit and traced through.  We get absolutely no notifications from fsevents when a broken symlink is created, not even on the containing directory.

I don't think we'll have a great way to detect this with fsevents; the best we can do is to periodically poll and crawl the filesystem to discover this type of change.

```
$ git diff root.c
diff --git a/root.c b/root.c
index 5ff74e1..5637384 100644
--- a/root.c
+++ b/root.c
@@ -2051,7 +2051,7 @@ static const struct watchman_hash_funcs root_funcs = {
 void watchman_watcher_init(void) {
   watched_roots = w_ht_new(4, &root_funcs);

-#if HAVE_FSEVENTS
+#if HAVE_FSEVENTS_NOPE
   watcher_ops = &fsevents_watcher;
 #elif defined(HAVE_PORT_CREATE)
   // We prefer portfs if you have both portfs and inotify on the assumption
```

can you try the above patch?  That will make watchman use kqueue directly.
You'll probably want to bump up your limits per https://facebook.github.io/watchman/docs/install.html#max-os-file-descriptor-limits

avsm:
According to an Apple Radar bug I filed, this should now be fixed in OSX 10.12 upwards:

> The original report on this issue, Bug ID 23193224, is closed.
> If you still see the issue in macOS 10.12, please file a new bug report.  The issue should be resolved there.

wez:
Confirmed fixed; I'll be pushing a commit to re-enable the tests for this situation on macOS and will close out this issue.  Thanks for letting me know!

wez:
[...]

PR #189, opened 2015-11-25
watchman: try to detect symlink changes more reliably on osx

wez:
Summary: Looking for a way to mitigate
https://github.com/facebook/watchman/issues/172 wherein fsevents won't
send us notifications about certain classes of changes involving
symlinks; the main one we're aware of is where a symlink is changed to
point to a file that doesn't (yet) exist on the filesystem.

The approach in this diff is to add a configuration option
(`recheck_dirs_interval_ms`) to specify how long we'll wait before we go
back and re-examine candidate files.

This parameter is checked after we've settled and defaults to the
trigger settle interval on mac, but is turned off for Linux.

Once we reach this point in time, we perform a query over the recently
changed file nodes and collect all dir nodes or the dir-node parents of
the files that have changed since we last checked for this category of
change.

If this list is empty then nothing has changed so we have no further
work to do.

Otherwise, we then search all symlink file nodes and add them to the
set, and we then re-examine this complete list of file nodes and
directories to try to discover changes that fsevents failed to inform us
of.

After this check we look to see how many nodes were fixed up; if there
were any fixed up nodes we'll re-set the settle to its base.

I found a couple of bugs while looking at this:

1. We didn't respect the recursive flag in `crawler` and would
unconditionally recursively crawl every item contained in a dir.

2. If we re-examined a file node that was already !file->exists and
realized that it still didn't exist, we'd observe a change for it
even though it logically could not have changed (it was deleted and
still is deleted).

3. YATOCTOU (Yet Another...) when we `lstat(2)`; we can traverse
symlinks in the path, so we add `w_lstat` which uses our strict
checking for the path.  This fixes a bug where we'd want to assess
a path that had previously been deleted when its parent transitioned
from a dir to a symlink and then we'd flip-flop between exists and
!exists.

4. I added an additional fstat in our opendir implementation on mac;
there is a TOCTOU-style issue where we can return a dir handle for
something that isn't actually a dir any more (for dir->symlink
transitions) and the deferred detection of the error condition is
suboptimal.  We now stand a greater chance of detecting and dealing with
this condition when we open the dir.

Test Plan: `make integration` has no regressions.  Then I repeatedly
checked out between two revisions in a large repo where there are a
large number of dir and symlink transitions:

```
$ hg co master ; rm .hg/watchman.state ; hg status --time
$ hg co arcpatch-XYZ ; rm .hg/watchman.state ; hg status --time
```

without this diff, the second step would consistently report 3 files
as modified in the `hg status` output.  A subsequent `hg status`
wouldn't show the same files due to the way that it records and
re-checks notable files.

To validate that we're not busy waiting or polling too aggressively:

```
$ watchman -p --server-encoding=json log-level debug | jq .log
```

is quiet once we've settled.

Activity monitor shows 0% CPU in this state.

### DIFF
--- a/opendir.c
+++ b/opendir.c
@@ -134,6 +134,46 @@ static int open_dir_as_fd_no_symlinks(const char *path, int flags) {
 #endif
 }
 
+// Like lstat, but strict about symlinks in any path component
+int w_lstat(const char *path, struct stat *st) {
+#ifdef HAVE_OPENAT
+  char *pathcopy = strdup(path);
+  char *slash;
+  int dir_fd, res, err;
+
+  if (!pathcopy) {
+    return -1;
+  }
+
+  slash = strrchr(pathcopy, '/');
+  if (!slash) {
+    free(pathcopy);
+    errno = EINVAL;
+    return -1;
+  }
+
+  *slash = '\0';
+  dir_fd = open_dir_as_fd_no_symlinks(pathcopy, O_NOFOLLOW|O_CLOEXEC);
+  if (dir_fd == -1) {
+    free(pathcopy);
+    errno = ENOTDIR;
+    return -1;
+  }
+
+  errno = 0;
+  res = fstatat(dir_fd, slash + 1, st, AT_SYMLINK_NOFOLLOW);
+  err = errno;
+
+  free(pathcopy);
+  close(dir_fd);
+
+  errno = err;
+  return res;
+#else
+  return lstat(path, st);  // tadaa!
+#endif
+}
+
 /* Opens a directory making sure it's not a symlink */
 DIR *opendir_nofollow(const char *path)
 {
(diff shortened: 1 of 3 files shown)


wez:
Adding @bhamiltoncx as an FYI: I think this may help with that nasty issue we've been trying to run down

bhamiltoncx:
Hmm, pretty interesting. I haven't seen any suggestion that any IDE editors
use dangling symlinks, but perhaps the conflict resolution process for SCMs
invoking vi or emacs would cause dangling symlinks to be created.

Can we confirm that?
2015年11月24日(火) 19:15 Wez Furlong <notifications@github.com>:

wez:
We're still looking for a repro for that other situation, so it's difficult to say at this stage :-/

wez:
Pushed as 7edcb2110ee0c9121f9b7013dd4ec819407e1afc

PR #192, opened 2015-12-12
watchman: improve handling of case insensitive filenames

wez:
TL;DR:

1. Halved memory requirements for case insensitive filesystem trees
2. Halved crawl time on mac
3. Handle directory case-only-renames more reliably
4. Disable symlink re-checking (7edcb2110ee0c9121f9b7013dd4ec819407e1afc) because it is too costly in our largest repo when not using a sparse checkout

Copy-n-paste from the commit summary:

    Summary: we hit an issue where a directory rename that changed only the
    case of the dir name got us confused.  It had a bad interaction with
    the recently added symlink sanity checking code; the problem manifested
    as watchman relatively busily ping-ponging between thinking that `DIR/a`
    and `dir/a` existed.

    The heart of this issue is similar to the one that we solved with
    symlinks; at any point, any path component could have changed case, but
    on a case insensitive filesystem we can't ask the system to operate in
    case sensitive mode for a given operation.

    The resolution is similar to our resolution of the symlink issue; as we
    walk down from the root, not only do we check for symlinks, but we now
    also check the canonical case of the path component.  If what we want
    to open doesn't match the canonical case, we report ENOENT.

    This gives us a pretty decent case-sensitive stat operation (modulo
    TOCTOU, but it's no worse than our existing symlink checking).

    In our stat_path/crawler code, we check for ENOENT/ENOTDIR on case
    insensitive filesystems and use that as a signal to look in the parent
    of the item that we tried to examine.

    This serves two purposes:

    1. It allows us to discover the new canonical name by walking the dir
       contents.
    2. We can garbage collect any other similarly case-changed items in that
       dir.

    A nice side effect of this is that we can remove the relatively complex
    lc_files hash stuff; our core data structures can all be case sensitive
    and we can leave the insensitive matching to the query engine.  This
    means that we can halve the memory used to build our filesystem tree
    representation on Mac (nice!) but does make it a bit more expensive to
    crawl the filesystem.

    So, looking at perf; implementing things strictly as described above
    adds 50% to the initial crawl time for one of our large repos (so 150%
    total time vs. prior implementation).

    The original intent of the code that walks down from the root was to
    avoid going out to am expensive remote filesystem if a symlink change
    surprised us.  Since that type of change is not common, and because
    we're happy to now perform case sensitive compares on the path vs. the
    canonical path, we can switch to using a system-dependent means for
    obtaining the canonical path from an opened file descriptor.

    This switch takes us from a negative perf impact to a positive perf
    impact; 50% of the prior implementation for crawling this same large
    repo.

### DIFF
--- a/opendir.c
+++ b/opendir.c
@@ -47,143 +47,293 @@ struct watchman_dir_handle {
   struct watchman_dir_ent ent;
 };
 
-/* Opens a directory, strictly prohibiting opening any symlinks
- * in any component of the path on systems that make this
- * reasonable and easy to do.
- * We do this by opening / and then walking down the tree one
- * directory component at a time.
- * We prefer to do this rather than using realpath() and comparing
- * names for two reasons:
- * 1. We should be slightly less prone to TOCTOU issues if an element
- *    of the path is changed to a symlink in the middle of our examination
- * 2. We can terminate as soon as we discover a symlink this way, whereas
- *    realpath() would walk all the way down before it could come back to
- *    us and allow us to detect the issue.
+#ifdef _WIN32
+static const char *w_basename(const char *path) {
+  const char *last = path + strlen(path) - 1;
+  while (last >= path) {
+    if (*last == '/' || *last == '\\') {
+      return last + 1;
+    }
+    --last;
+  }
+  return NULL;
+}
+#endif
+
+static bool is_basename_canonical_case(const char *path) {
+#ifdef __APPLE__
+  struct attrlist attrlist;
+  struct {
+    uint32_t len;
+    attrreference_t ref;
+    char canonical_name[WATCHMAN_NAME_MAX];
+  } vomit;
+  char *name;
+  const char *base = strrchr(path, '/') + 1;
+
+  memset(&attrlist, 0, sizeof(attrlist));
+  attrlist.bitmapcount = ATTR_BIT_MAP_COUNT;
+  attrlist.commonattr = ATTR_CMN_NAME;
+
+  if (getattrlist(path, &attrlist, &vomit,
+        sizeof(vomit), FSOPT_NOFOLLOW) == -1) {
+    w_log(W_LOG_ERR, "getattrlist(%s) failed: %s\n",
+        path, strerror(errno));
+    return false;
+  }
+
+  name = ((char*)&vomit.ref) + vomit.ref.attr_dataoffset;
+  return strcmp(name, base) == 0;
+#elif defined(_WIN32)
+  WCHAR long_buf[WATCHMAN_NAME_MAX];
+  WCHAR *wpath = w_utf8_to_win_unc(path, -1);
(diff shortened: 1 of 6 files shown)


wez:
Bah, I'm going to have to dust off my win32 environment to tidy this up

opendir.c
```diff
@@ -47,143 +47,294 @@ struct watchman_dir_handle {
...
+  WCHAR final_buf[WATCHMAN_NAME_MAX];
+  DWORD len, err;
+  DWORD nchars = sizeof(final_buf) / sizeof(WCHAR);
+
+  if (h == INVALID_HANDLE_VALUE) {
+    return -1;
+  }
+
+  len = GetFinalPathNameByHandleW(h, final_buf, nchars,
+                                  FILE_NAME_NORMALIZED | VOLUME_NAME_DOS);
```
bhamiltoncx: `VOLUME_NAME_DOS` is the default, no need to pass it.

opendir.c
```diff
@@ -47,143 +47,294 @@ struct watchman_dir_handle {
...
+    char *utf8 = w_win_unc_to_utf8(final_buf, len);
+    if (!utf8) {
       return -1;
     }
 
-    // walk down to next level
-    elem = next;
-    close(fd);
-    fd = next_fd;
+    uint32_t utf_len = u32_strlen(utf8);
```
bhamiltoncx: Just a note: `u32_strlen()` is a terrible name for this function, because there is such a thing as UTF-32, and it's not clear from the name that this isn't returning the length of the UTF-8 string when encoded in UTF-32.

(Reading the implementation makes it clear it's actually `uint32_t`.)

opendir.c
```diff
@@ -47,143 +47,294 @@ struct watchman_dir_handle {
...
+  err = errno;
+#else
+  res = lstat(path, st);
   err = errno;
+#endif
 
-  free(pathcopy);
-  close(dir_fd);
+  if (res == 0 && !case_sensitive && !is_basename_canonical_case(path)) {
+    err = ENOENT;
```
bhamiltoncx: Style: All the other blocks assign to `res` before `err`, why not do it here too?

root.c
```diff
@@ -854,13 +790,13 @@ static void stat_path(w_root_t *root,
     /* it's not there, update our state */
     if (dir_ent) {
       w_root_mark_deleted(root, dir_ent, now, true);
-      w_log(W_LOG_DBG, "lstat(%s) -> %s so stopping watch on %s\n",
-          path, strerror(err), dir_ent->path->buf);
+      w_log(W_LOG_DBG, "w_lstat(%s) -> %s so stopping watch on %.*s\n", path,
```
bhamiltoncx: Nice fix. :)

bhamiltoncx:
Looks good overall, didn't see any leaks.

PR #196, opened 2016-01-04
pybser: allow st_size in __getattr__

quark-zju:
Fixes `AttributeError: _BunserDict has no attribute st_size` when bser.c is not
compiled. Traceback:

    File "hg/mercurial/dirstate.py", line 1101, in status
      ((size != st.st_size and size != st.st_size & _rangemask)
    File "hgwatchman/hgwatchman/pywatchman/pybser.py", line 245, in __getattr__
      raise AttributeError('_BunserDict has no attribute %s' % name)

### DIFF
--- a/python/pywatchman/pybser.py
+++ b/python/pywatchman/pybser.py
@@ -239,10 +239,7 @@ class _BunserDict(object):
         self._values = values
 
     def __getattr__(self, name):
-        try:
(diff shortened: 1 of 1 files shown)


wez:
Pushed as  https://github.com/facebook/watchman/commit/2ca4c24663627219bfb6fc14882b13a8c057da7d
Thanks!

PR #199, opened 2016-01-13
Guard against receiving data after the socket was closed.

cpojer:
I'm using this implicitly on www and it throws every time. For some reason I'm receiving data after the socket was closed (wtf?). Logging the buffer gives me this: `<Buffer 00 01 04 48 01 01 03 07 02 03 07 76 65 72 73 69 6f 6e 02 03 05 34 2e 31 2e 30 02 03 05 63 6c 6f 63 6b 02 03 1c 63 3a 31 34 35 32 35 34 36 33 31 34 3a ... >`

Is this a reasonable guard?

@wez @amasad

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -118,7 +118,9 @@ Client.prototype.connect = function() {
       self.emit('error', err);
     });
     self.socket.on('data', function(buf) {
-      self.bunser.append(buf);
+      if (self.bunser) {
+        self.bunser.append(buf);
+      }
     });
     self.socket.on('end', function() {
       self.socket = null;


wez:
That sounds like it will fix the throw, but I wonder if we have a bigger problem.  How is the connection ended?  What commands are you sending to watchman; are they subscription results or something else?

I'm wondering whether we need to defer emitting our own `end` signal until we've processed the pending data in the socket, or whether we need to explicitly `flush` or kill off that socket object in the `end` handler.

cpojer:
I don't really know much about watchmen but this is coming from sane when using it together with haste2 ( https://github.com/facebook/node-haste ). Not entirely sure where it is coming from as there is a Promise chain involved and the stack trace isn't helpful. See https://phabricator.fb.com/P56086717

wez:
Is the socket ending due to being CTRL-C'd or because it is being closed by code?

cpojer:
Closed by code in this case.

wez:
OK, I think this is probably the best/simplest thing.  I'll merge it in; thanks!

wez:
Pushed as acba4edba87193c7ba73f8b6afdc75d5d2823578 and dumped the package version in npm to 1.8.0

cpojer:
@zertosh mentioned that this fixes a lot of issues with Nuclide?

zertosh:
We spin up separate node processes for the nuclide file search - which uses fb-watchman. I see a lot of uncaught exceptions because of fb-watchman. I don't know if it's exactly this line. The subprocess stack traces are pretty poor. But I'll report back if I see a decrease in the number of exceptions.

zertosh:
FYI, a better fix is to remove the `data` handler after `end`

PR #208, opened 2016-02-03
inotify: don't unconditionally set the recursive flag

wez:
Summary: durham noticed that calls to `clock` were taking 6 seconds to
return during some rebase profiling with hgwatchman.  Looking deeper,
we're doing a lot of recursive walking.  I don't recall why we were
setting recursive to true here, but turning it off seems to have no
negative side effects; the watchman integration tests pass, as do
the mercurial tests.

Test plan: ran hgwatchman + hg tests, `make integration` passes,
corrected the slow rebase behavior.  Let's see what Travis CI makes of
it.


wez:
CI failures are unrelated (also manifest in master), #209 is up to help understand them

PR #209, opened 2016-02-03
travis: restore dumping error logs on test failure

wez:
Trying to understand a test failure on Travis and noticed that our logs
aren't where they used to be


wez:
pushed

PR #211, opened 2016-02-14
Use strsignal() instead of sys_siglist

Vaelatern:
This fixes watchman for building with non-glibc
libcs, such as musl.

No longer relying on implementation details seems
like the right direction.

https://travis-ci.org/voidlinux/void-packages/builds/109134106


wez:
Thanks for digging in!

I am mildly concerned that this makes our signal handler less async-signal safe than it is already.  Right now we stand a pretty decent chance of successfully logging something useful to the log stream before we try anything fancy like logging a stack trace.

`strsignal` allocates memory on some (all?) platforms and would potentially make things harder to diagnose in the case of heap corruption.  This is the main reason why I chose not to use strsignal.  Regarding your comment about relying on implementation details, `sys_siglist` is a documented part of the API and ABI on the systems that we support.

If musl doesn't provide a sys_siglist alternative, please amend this PR to make the use of strsignal dependent on either a preprocessor feature test or a configure check; thanks!

Vaelatern:
I'll come back to this PR and look at a conditional compilation option.

wez:
Closing this out due to it being stale; looking forward to the next iteration of it when you have time.  Thanks!

PR #215, opened 2016-02-18
fix inotify race condition

wez:
Summary: This fixes an issue that we found on linux after introducing
222447241292c10d0a4eb49ec8659515456d1cb6

What happens is that there is a race condition in the case that a
directory tree is non-atomically replaced and rebuilt; we end up missing
the information about the tree being rebuilt.

This diff adds an integration test to reproduce the problem scenario and
fixes it by conditionally setting the recursive flag in the handful of
cases that make sense.

Test Plan: `make integration`

Reviewers: sid0

CC:

Task ID: #9981085

Blame Rev:


wez:
pushed as 46ed00701b7d0a722a8da350464f945bdee59691

PR #219, opened 2016-02-25
Make pywatchman run on Python 3

Qwait:
Python 3 doesn't support the long character (L), int and long types have been merged in Python 3. "except Exception, e" has been removed in Python 3.

It is impossible to test pywatchman with Python 3.4 with TravisCI, since Python 3.4 is not natively supported on TravisCI's Ubuntu 12.04.5, but this a separate issue all together.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
     packages:
       - php5-cli
       - python2.7
+      - python3.4
 #      - ruby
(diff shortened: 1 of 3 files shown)


wez:
Looks good, thanks!  I'll `hub am` this a bit later today

wez:
Squashed and pushed, thanks!

PR #221, opened 2016-02-29
Improve diagnostics and behavior when the root lock is held for a long time

wez:
We can now annotate the root with the reason that the lock is held, and use an alternative locking function that allows specifying a timeout.

Between these and #220 we should be able to understand a bit more about the issue that @int3 has been tracking.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,6 +30,7 @@ watchman_SOURCES = \
 	stream.c        \
 	stream_stdout.c \
 	stream_unix.c   \
+	timedlock.c     \
 	cmds/find.c     \
 	cmds/info.c     \
 	cmds/log.c      \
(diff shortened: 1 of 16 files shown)


wez:
re: 10 ms.  Yeah, it may be a little short.  In practice, we just gave up the lock after processing updates.  I think I'll bump this up to 100ms so that it is bounded and short, but not too short. (we can always come back and dispatch on a subsequent go around).  I'll make this configurable in case we later regret this choice.

The idea for this PR is that we sometimes see hh_server queries timeout after 2 minutes with no evidence of why in the logs (it's not sync_timeout) so the implication is that we're stuck somewhere.  The idea here is that we can set an upper bound on the time we're willing to wait for the cases where we can safely fail out early (such as queries) and have that report reasonably back to the client so that we have more context than just a hard socket timeout.

#220 has some similar / related changes that allow more general sampling of where things are running slow

wez:
Pushed as sequence ending with 72f42bcbf333bc06d777a5f6eef7539da603ea80

Issue #222, opened 2016-03-01
Expand concept of authentication


wez:
We have an internal use case where it is desirable for a process that runs as a different user from a user-owned watchman process to consume information from watchman.  The idea is that we don't want to have to commit multiple processes to watching the same tree(s).

In this scenario, the plan is to change the group of the unix socket and make it group writable so that this other process can connect.

This raises a security concern in that some of the watchman commands allow registering commands to be run as the owning user of the service.

To facilitate this use case, these things are needed:

1. Expand the various watchman clients so that they use SO_PEERCRED to transmit their credentials
2. Add a `CMD_REQUIRE_AUTH_OWNER` flag for the command dispatcher to have it check that the credentials of the client exactly match the owner and fail automatically if they don't.  We'd definitely want to set this for `trigger` and probably also for the `watch` and `watch-project` commands.
3. Add `sock_path_group` and `sock_path_mode` global configuration options that cause the listener to change the group and explicitly chmod the unix socket path after it has started up the listener
4. Add a global `require_unix_credentials` boolean that controls whether the listener will accept any commands from a client that hasn't passed in credentials.

The upshot of this is that it means that older clients that don't pass credentials won't be able to establish triggers; this is a behavior change, but it sounds reasonable.

Should also take a look at #204 and the proposed authenticate command in the protocol and see if that changes the outlook here.

wez:
With Sid's recent changes, I think we're done here.

PR #223, opened 2016-03-02
Add concept of session owner, restrict command execution

wez:
Refs: https://github.com/facebook/watchman/issues/222

This adds an interface to our stream layer that allows us to check
whether the peer matches the owner of our process.  We do not
currently implement this for Windows.

We record this flag in the client structure.

The command dispatcher will by default refuse to allow any commands to
run if the peer is not the owner.

A command can be registered with the `CMD_ALLOW_ANY_USER` to whitelist
that command from this blanket dispatcher check.

The model we're using to decide what is white listed is this:

* If you are not the owner, you can inspect any of the current
  state in the watchman instance.
* You may not change that state; initiating new watches, deleting
  watches or registering triggers are prohibited

`watch` and `watch-project` are white listed because they are an
important part of initializing a client; well behaved clients will
issue `watch-project` to resolve the watched path and directory
offset they should use for their session.

For that reason, we add a check for client==owner when we resolve a
root; if the `auto_create` flag is true and you are not the owner,
we turn the flag off.  We'll also expand the error message to provide
more context.

Test Plan: included a test case, but also manually:

```
$ sudo watchman -U /opt/facebook/watchman/var/run/watchman/wez-state/sock --no-spawn watch list
{
    "version": "4.5.0",
    "error": "unable to resolve root list: path \"list\" must be absolute (this may be because you are not the process owner)"
}
$ sudo watchman -U /opt/facebook/watchman/var/run/watchman/wez-state/sock --no-spawn watch-list
{
    "version": "4.5.0",
    "roots": [
        "/Users/wez/fbsource",
        "/Users/wez/fb/watchman"
    ]
}
$ sudo watchman -U /opt/facebook/watchman/var/run/watchman/wez-state/sock --no-spawn watch-del /Users/wez/fbsource
{
    "version": "4.5.0",
    "error": "you must be the process owner to execute 'watch-del'"
}
$ sudo watchman -U /opt/facebook/watchman/var/run/watchman/wez-state/sock --no-spawn watch /Users/wez/fbsource
{
    "version": "4.5.0",
    "watch": "/Users/wez/fbsource",
    "watcher": "fsevents"
}
```

### DIFF
--- a/cmds/debug.c
+++ b/cmds/debug.c
@@ -128,5 +128,18 @@ static void cmd_debug_poison(struct watchman_client *client, json_t *args)
 }
 W_CMD_REG("debug-poison", cmd_debug_poison, CMD_DAEMON, w_cmd_realpath_root)
 
+static void cmd_debug_drop_privs(struct watchman_client *client, json_t *args)
+{
+  json_t *resp;
+
+  unused_parameter(args);
+  client->client_is_owner = false;
+
+  resp = make_response();
+  set_prop(resp, "owner", json_boolean(client->client_is_owner));
+  send_and_dispose_response(client, resp);
+}
+W_CMD_REG("debug-drop-privs", cmd_debug_drop_privs, CMD_DAEMON, NULL);
+
 /* vim:ts=2:sw=2:et:
  */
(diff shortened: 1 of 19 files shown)


wez:
Pushed as d22ef3b7d8ca7a7720608727944464c232821f8b and 189a4fac4bdf5730f24869c4012def0fc49af0fb

PR #224, opened 2016-03-02
Update the license metadata.

mjpieters:
All source files already use the BSD 3-clause text, this change just reflects
the correct license in the package metadata.


wez:
Pushed as 150f1fb4eff94c0248b54154fb356f768a131d60, thanks!

PR #231, opened 2016-03-09
Flush output for each line

### LINKED ISSUE #230: watchman-wait does't work with xargs
This works:

```
watchman-wait -m 0 .
# in another window `touch me`
me
```

This doesn't work:

```
watchman-wait -m 0 . | xargs echo
# in another window `touch me`
# nothing is printed
```

```
$ watchman --version
4.4.0
$ sw_vers -productVersion
10.11.3
```

ptarjan:
When piping to another process, linux will buffer the output. This breaks line-based processing like most people would do with `xargs` for `watchman-wait`.

Fixes #230

### DIFF
--- a/python/bin/watchman-wait
+++ b/python/bin/watchman-wait
@@ -166,6 +166,7 @@ Perhaps you should use the --pattern option?""" % (path, self.path), file=sys.st
                     for fname in args.fields:
                         out.append(self.formatField(fname, f[fname]))
                 print(args.separator.join(out))
+                sys.stdout.flush()
                 total_events = total_events + 1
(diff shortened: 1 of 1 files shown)


ptarjan:
Is CircleCI supposed to pass?

wez:
It should... be I only just added this past week.  Looks like they aren't reachable right now anyway :-/

wez:
Thanks Paul!

Applied via `hub am https://github.com/facebook/watchman/pull/231` and pushed as b203f99.

Things are good here; hope you're enjoying your new gig!

ptarjan:
Enjoying it. But missing seeing you :) (And Ruby is FAR worse than Hack...)

PR #242, opened 2016-04-17
Use strsignal() if sys_siglist[] is unavailable

Vaelatern:
This is a problem on muslc, which tries to be as minimal
as possible. A side effect of this is implementation details
not seen as integral to a libc, sometimes change, and sys_siglist was one
of those changes.

Previous PR #211 which was abandoned due to lack of time.

A sample build of watchman on multiple platforms and libcs can be seen [here](https://travis-ci.org/Vaelatern/void-packages/builds/123763483).

This patch is also active in [Void Linux's version of this package.](https://github.com/voidlinux/void-packages/blob/master/srcpkgs/watchman/patches/musl_configure.patch)


Vaelatern:
AppVeyor build error seems entirely unrelated.

wez:
Thanks for coming back to this!

Would you mind factoring out the ifdefs a bit?

Something like:

```
#ifdef HAVE_SYS_SIGLIST
# define w_strsignal(x)  sys_siglist[x]
#else
# define w_strsignal(x)  strsignal(x)
#endif
```

and then just using `w_strsignal` in the main body of the code, so that it is slightly less messy?

configure.ac
```diff
@@ -251,6 +251,8 @@ AC_CHECK_HEADERS(CoreServices/CoreServices.h, [
 ])
 AC_CHECK_FUNCS(backtrace backtrace_symbols backtrace_symbols_fd)
 
+AC_CHECK_FUNCS_ONCE(sys_siglist)
```
wez: does this need to be `_ONCE`?

Vaelatern: `_ONCE` takes care of more of the problems such as what to define, printing output, and has a simpler signature.

Vaelatern: But it seems I could remove `_ONCE` and have no effect.
https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Functions.html

Vaelatern:
Changes effected.

Vaelatern:
```
FAIL tests/integration/watch-del-all.php (0.292s)
Traceback (most recent call last):
  File "/home/ubuntu/.pyenv/versions/2.7.5/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/home/ubuntu/watchman/tests/integration/WatchmanTapTests.py", line 89, in runTest
    self.fail(msg)
  File "/home/ubuntu/.pyenv/versions/2.7.5/lib/python2.7/unittest/case.py", line 412, in fail
    raise self.failureException(msg)
AssertionError: not ok 11 - watch-del-all.php:31: file lists are not equal at line 31 in file watch-del-all.php

file lists are not equal at line 31 in file watch-del-all.php
```

wez:
I've scheduled a rebuild in CircleCI, it's almost certainly unrelated to the this PRI

Vaelatern:
At the least not deterministically related to this PR. Probably not related at all.

wez:
Merged, thanks!

PR #243, opened 2016-04-17
Use more reliably present poll.h

Vaelatern:
Void Linux currently has a blanket replacement in place: `sed -i 's,sys/poll.h,poll.h,g' *.[ch]`.

Since that line was written (at version 3.3.0), it seems `sys/poll.h` now only gets referenced from `watchman.h`.

That said, `poll.h` is present on more systems, even if it only contains the line `#include <sys/poll.h>`.

Unless this is not true for a watchman supported platform, this change should work. If this breaks your build, I can add a configure check.

### DIFF
--- a/watchman.h
+++ b/watchman.h
@@ -55,7 +55,7 @@ extern "C" {
 # define O_CLOEXEC 0
 #endif
 #ifndef _WIN32
-#include <sys/poll.h>
+#include <poll.h>
 #include <sys/wait.h>
 #endif
 #ifdef HAVE_PCRE_H


wez:
We don't have CI for Solaris, but can you check portability for that platform also?

Vaelatern:
I am asking a fellow with access to a Solaris box, but until then, this suggests it exists. https://blogs.oracle.com/vlad/entry/poll_and_pollhup_in_solaris

wez:
That's good enough; merged, thanks!

PR #250, opened 2016-05-22
Use a prefix tree for the ignore list

wez:
When we first started out the thought was that very few directories would be ignored, but as we have progressed various internal projects have added a surprising number of ignore dirs in our internal repos.

We now find that our simplistic ignore checker is underperforming, particularly in the ignore-miss case as we have to compare N string prefixes before realizing that there is no match.

This pull request is a short series of diffs that:

1. Imports an Adaptive Radix Tree implementation that we will use as accelerator for string prefix matching
2. Adds unit tests for the underlying ignore checking functions that are used only on OS X and Windows
3. Augments the Radix Tree implementation with a function that can find the longest matching prefix
4. Benchmarks the prefix matching tree implementation against the prior simplistic solution
5. Replaces the old solution with the prefix matching tree implementation.

The benchmarks show that, for an `ignore_dirs` list of equivalent size and content as one used in a real repo, that the ignore-miss case is now approximately 11x faster and the ignore-hit case is approximately 2x faster.  The latter can also be up to 11x faster if the ignore rule that is hit is towards the end of the ignore list.

The latency of this routine is important to minimize the risk of an overflow event.


facebook-github-bot-0:

facebook-github-bot-6:

facebook-github-bot-6:

sid0:
Looks like the AppVeyor build is genuinely broken.

facebook-github-bot-5:

facebook-github-bot-1:

wez:
Yeah, I'm iterating away on the windows build via appveyor :-p

facebook-github-bot-4:

facebook-github-bot-1:

facebook-github-bot-9:

facebook-github-bot-9:

facebook-github-bot-1:

facebook-github-bot-0:

facebook-github-bot-3:

facebook-github-bot-1:

facebook-github-bot-5:

facebook-github-bot-9:

wez:
Windows build seems ok (albeit with that dir move failure), but this test is a bit slower than I'd like; this runs in about 1.5 seconds on my laptop.  Circle CI on Linux has this running at around 4s.

```
PASS tests\ignore.exe (7.392s)
```

facebook-github-bot-1:

facebook-github-bot-9:

facebook-github-bot-2:

wez:
Re: windows perf, it's just pretty awful.  I don't know how much of this is appveyor vs. compiler flags, but at least this PR improves the state of things.

The output below compares the simplistic implementation with the libart implementation:

```
ok 34 - builda/hello expected=0 actual=0
# all_ignores_iter: took 15.000s
# all_ignores_tree: took 5.000s
# no_ignores_iter: took 15.188s
# no_ignores_tree: took 3.001s
not ok 35 - just want to see the output
#     Failed test (tests\ignore_test.c:main() at line 203)
```

ignore.c
```diff
@@ -0,0 +1,117 @@
+/* Copyright 2016-present Facebook, Inc.
+ * Licensed under the Apache License, Version 2.0 */
+
+#include "watchman.h"
+
+// The path and everything below it is ignored.
+#define kFullIgnore  (void*)0x1
```
sid0: Our standard would be something like `FULL_IGNORE` and `VCS_IGNORE`

ignore.c
```diff
@@ -0,0 +1,117 @@
...
+
+  art_insert(&ignore->tree, (const unsigned char *)path->buf, path->len,
+             is_vcs_ignore ? kVCSIgnore : kFullIgnore);
+}
+
+bool w_ignore_check(struct watchman_ignore *ignore, const char *path,
+                    uint32_t pathlen) {
+  const char *skip_prefix;
+  uint32_t len;
+  art_leaf *leaf = art_longest_match(&ignore->tree, (const unsigned char *)path,
```
sid0: `const unsigned char`? Really? :/

ignore.c
```diff
@@ -0,0 +1,117 @@
...
+  // Our input string was longer than the leaf key string.
+  // We need to ensure that we observe a directory separator at the
+  // character after the common prefix, otherwise we may be falsely
+  // matching a sibling entry.
+  skip_prefix = path + leaf->key_len;
+  len = pathlen - leaf->key_len;
+
+  if (*skip_prefix != WATCHMAN_DIR_SEP
+#ifdef _WIN32
+      && *skip_prefix != '/'
```
sid0: Hmm, why is this separate check needed? Is it just because people on Windows can specify ignores with both `/` and `\`? Probably worth adding a comment explaining that.

ignore.c
```diff
@@ -0,0 +1,117 @@
...
+
+  // If we find a '/' from this point, we are ignoring this path.
+  return memchr(skip_prefix, WATCHMAN_DIR_SEP, len) != NULL
+#ifdef _WIN32
+         || memchr(skip_prefix, '/', len) != NULL
+#endif
+      ;
+}
+
+void w_ignore_destroy(struct watchman_ignore *ignore) {
```
sid0: Our usual standard would be `w_ignore_free`

sid0:
Looks great to me, modulo the couple of comments I had. Thanks!

ignore.c
wez: yeah, both are possible on windows.  I'll add a coment

ignore.c
wez: for me, `foo_free` implies that it will call `free(foo)` at the end, whereas `foo_destroy` just tears down the internals of a foo without freeing the foo itself

sid0: Hmm, can we call `free` here too? That'll make stuff less confusing overall I think. (Or are we storing this stuff inline in another `struct`?)

wez: On Thu, May 26, 2016 at 11:27:40AM -0700, Siddharth Agarwal wrote:
> Hmm, can we call free here too? That'll make stuff less confusing overall I
> think. (Or are we storing this stuff inline in another struct?)

It's embedded in the w_root_t so we must not free it here.

facebook-github-bot-2:

PR #251, opened 2016-05-24
Critbit based prefix tree

wez:
Similar to https://github.com/facebook/watchman/pull/250 but this one adds and uses @sid0's critbit tree implementation.

The perf isn't as good as I'd like; I think this is because we can't compare strings until we've walked the tree, and we may have to backtrace:

compare the benchmark against libart.

This is libart:

```
$ ./tests/ignore.t
...
 # all_ignores_tree: took 0.974s
 # no_ignores_tree: took 0.427s
 # ELAPSED: 1597ms
```

This is critbit:

```
 # all_ignores_tree: took 1.980s
 # no_ignores_tree: took 1.425s
 # ELAPSED: 3603ms
```

Was hoping they'd be closer :-/


sid0:
Ah well, guess it was worth a shot :)

wez:
Abandoning this in favor of the libart approach

PR #253, opened 2016-05-25
Support for symbolic link target handling

anirbanr-fb:
Summary: Add a field for symlink_target in watchman file.
Add a query field to return the symlink_target for the file.

When symbolic link changes, update the target metadata.

Test Plan: Added tests for symlink. Some of these tests are
not run on Darwin and NT, because of platform specific behavior.

1. test for basic symlink targets.
2. invalid symlinks.
3. deletion of symlinks.
4. transition of symlink to regular file or directory

Reviewers: wez

Subscribers:

Tasks:

Blame Revision:

### DIFF
--- a/root.c
+++ b/root.c
@@ -902,6 +902,7 @@ static void stat_path(w_root_t *root,
 
     memcpy(&file->stat, &st, sizeof(file->stat));
 
+#ifndef _WIN32
     // check for symbolic link
     if (S_ISLNK(st.mode)) {
       char link_target_path[WATCHMAN_NAME_MAX];
@@ -928,6 +929,7 @@ static void stat_path(w_root_t *root,
       w_string_delref(file->symlink_target);
       file->symlink_target = NULL;
     }
+#endif
 
     if (S_ISDIR(st.mode)) {
       if (dir_ent == NULL) {
--- a/root.c
+++ b/root.c
@@ -902,6 +902,33 @@ static void stat_path(w_root_t *root,
 
     memcpy(&file->stat, &st, sizeof(file->stat));
 
+    // check for symbolic link
+    if (S_ISLNK(st.mode)) {
+      char link_target_path[WATCHMAN_NAME_MAX];
+      ssize_t tlen = 0;
+
+      tlen = readlink(path, link_target_path, WATCHMAN_NAME_MAX);
+      if (tlen < 0 || tlen >= WATCHMAN_NAME_MAX) {
+        w_log(W_LOG_ERR,
+            "readlink(%s) errno=%d tlen=%d\n", path, errno, (int)tlen);
+
+        // not a proper symbolic link. NULL out symlink_target
+        if (file->symlink_target) {
+          w_string_delref(file->symlink_target);
+          file->symlink_target = NULL;
+        }
+      } else {
+        if (file->symlink_target) {
+          w_string_delref(file->symlink_target);
+        }
+
+        file->symlink_target = w_string_new_len(link_target_path, tlen);
+      }
+    } else if (file->symlink_target) {
+      w_string_delref(file->symlink_target);
+      file->symlink_target = NULL;
+    }
+
     if (S_ISDIR(st.mode)) {
       if (dir_ent == NULL) {
         recursive = true;
@@ -1561,6 +1588,9 @@ static void free_file_node(w_root_t *root, struct watchman_file *file)
 {
   root->watcher_ops->file_free(file);
   w_string_delref(file->name);
+  if (file->symlink_target) {
+    w_string_delref(file->symlink_target);
+  }
   free(file);
 }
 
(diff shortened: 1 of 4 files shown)


facebook-github-bot-4:
@anirbanr-fb updated the pull request.

facebook-github-bot-1:

root.c
```diff
@@ -902,6 +902,35 @@ static void stat_path(w_root_t *root,
 
     memcpy(&file->stat, &st, sizeof(file->stat));
 
+#ifndef _WIN32
+    // check for symbolic link
+    if (S_ISLNK(st.mode)) {
+      char link_target_path[WATCHMAN_NAME_MAX];
+      ssize_t tlen = 0;
+
```
wez: let's put this bit up here:

```
 if (file->symlink_target) {
       w_string_delref(file->symlink_target);
       file->symlink_target = NULL;
 }
```

so that we don't need to have three other places that we do the same thing; we can just focus on setting it in the success case.

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+
+        expr = {
+            "expression": ["name", "111"],
+            "fields": ["name", "symlink_target"]
+        }
+
+        res1 = self.watchmanCommand('query', root, expr)
+
+        files1 = res1['files']
+        self.assertTrue(files1 and os.path.basename(files1[0]['symlink_target']) == '222')
```
wez: prefer to do:

```
self.assertEquals('222', os.path.basename(files1[0]['symlink_target']))
```

so that the test assertion will show what the actual value was if it ever fails

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+
+        os.unlink(os.path.join(root, '111'))
+        os.symlink(os.path.join(root, '333'), os.path.join(root, '111'))
+
+        res2 = self.watchmanCommand('query', root, expr)
+        files2 = res2['files']
+        #print files2
```
wez: remove this commented out line

tests/integration/test_symlink.py
wez: I would just use `files['files'][0]` here rather than a local named `files1`, or do:

```
files = files['files']
```

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+        self.assertTrue(files2 and os.path.basename(files2[0]['symlink_target']) == '333')
+
+    # test to see that invalid symbolic link
+    # updates are picked up by the symlink_target field
+    @unittest.skipIf(sys.platform == 'darwin' or os.name == 'nt', 'mac os or win')
```
wez: add a comment:

> Skipping this test on mac due to the dangling symlink bug in fsevents

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+    def test_delSymlink(self):
+        root = self.mkdtemp()
+        self.touchRelative(root, '222')
+        os.symlink(os.path.join(root, '222'), os.path.join(root, '111'))
+
+        self.watchmanCommand('watch', root)
+        self.assertFileList(root, ['111', '222' ])
+
+        # Create a cursor for this state
+        self.watchmanCommand('since', root, 'n:foo')
```
wez: don't need a cursor here; just use a name query to match it

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+
+        expr = {
+            "fields": ["name", "symlink_target"],
+            "since" : 'n:foo',
+        }
+
+        res = self.watchmanCommand('query', root, expr)
+        files = res['files']
+        #print files
+        self.assertTrue(files and files[0]['name']  == '111'
```
wez: assertEquals

tests/integration/test_symlink.py
```diff
@@ -0,0 +1,137 @@
...
+
+        expr = {
+            "expression": ["name", "111"],
+            "fields": ["name", "symlink_target"]
+        }
+
+        res1 = self.watchmanCommand('query', root, expr)
+
+        files1 = res1['files']
+        self.assertTrue(files1 and os.path.basename(files1[0]['symlink_target']) == '222')
```
wez: use assertEquals for all of the equality assertions in this file

anirbanr-fb:
couple of assertTrue's remain. delSymLink unit test still has a since clause.

facebook-github-bot-7:
@anirbanr-fb updated the pull request.

facebook-github-bot-9:
@anirbanr-fb updated the pull request.

facebook-github-bot-3:
@anirbanr-fb updated the pull request.

wez:
Squashed and pushed as 2c32d252a15216df35928244c1bd03604ae8fb7f
Thanks!

PR #257, opened 2016-06-01
allow group/others to read information from watchman

sid0:
This is still a WIP, will finish off tomorrow

### DIFF
--- a/main.c
+++ b/main.c
@@ -11,6 +11,7 @@ static enum w_pdu_type server_pdu = is_bser;
 static enum w_pdu_type output_pdu = is_json_pretty;
 static char *server_encoding = NULL;
 static char *output_encoding = NULL;
+static char *test_state_dir = NULL;
 static char *sock_name = NULL;
 char *log_name = NULL;
 #ifdef USE_GIMLI
@@ -457,7 +458,7 @@ static void compute_file_name(char **strp,
     /* We'll put our various artifacts in a user specific dir
      * within the state dir location */
     char *state_dir = NULL;
-    const char *state_parent =
+    const char *state_parent = test_state_dir ? test_state_dir :
 #ifdef WATCHMAN_STATE_DIR
           WATCHMAN_STATE_DIR
 #else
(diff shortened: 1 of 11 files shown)


facebook-github-bot-3:
By analyzing the blame information on this pull request, we identified @wez and @Qwait to be potential reviewers.

facebook-github-bot-4:
@sid0 updated the pull request.

sid0:
This is now ready for review.

facebook-github-bot-0:
@sid0 updated the pull request.

facebook-github-bot-7:
@sid0 updated the pull request.

facebook-github-bot-6:
@sid0 updated the pull request.

main.c
```diff
@@ -520,6 +521,19 @@ static void compute_file_name(char **strp,
         goto bail;
       }
 
+      // Depending on group and world accessibility, change permissions on the
+      // directory. We can't leave the directory open and set permissions on the
+      // socket because not all POSIX systems respect permissions on UNIX domain
+      // sockets, but all POSIX systems respect permissions on the containing
+      // directory.
+      w_log(W_LOG_DBG, "Setting permissions on state dir to %#o", dir_perms);
```
wez: does this octal formatting work on windows and solaris too?

main.c
```diff
@@ -484,6 +484,8 @@ static void compute_file_name(char **strp,
       int dir_fd;
       int ret = 0;
       uid_t euid = geteuid();
+      const char *sock_group_name = cfg_get_string(NULL, "sock_group", NULL);
```
wez: I can see it being potentially useful to allow this to also be the gid (eg: we have a consistent gid for a role but deploy across systems where that group has a different name in the local groups file), but we can tackle that in the future.

tests/integration/WatchmanInstance.py
```diff
@@ -52,7 +52,8 @@ class _Instance(object):
     # overridden global configuration file; you may pass that
     # in to the constructor
 
-    def __init__(self, config={}):
+    def __init__(self, config={}, start_timeout=1.0):
```
wez: the default timeout went from 10 down to 1; intentional?

sid0: The default timeout was always 1 second -- 10 * 0.1 = 1.0

wez:
Looks good to me!  A couple of comments inline, but nothing blocking.  I think I reviewed this in reverse order (went from top to bottom in the list that GH showed me), but I think there are a couple of diffs near the bottom of the stack that may not be in the correct relative order; working checking with a `for-each-stacked make integration` run

sid0:
> top to bottom in the list that GH showed me

That should be earlier in the stack to later in the stack.

main.c
sid0: I'll switch this to `0%o`

main.c
sid0: I'll add a TODO

sid0:
Pushed as 63000aed37643be385258aa0aac60748f5d11e2c.

PR #258, opened 2016-06-01
Optimize pending list processing

wez:
This is a series of diffs to improve our handling of pending list processing.

There are a couple of diffs that are simple refactorings to libart, but there are two that bear special scrutiny:

* `libart: better defined behavior when a full prefix key is encountered`
* `use art to optimize pending list construction`

The former is necessary to avoid being forced to make a copy and NUL-terminate all input strings for both lookup, prefix iteration and insertion.

The latter is really the main component of this pull request.

In the included benchmark we can reduce a bloated pending list full of 260k pending lstat operations down to a dozen or so recursive walks.  This is advantageous because it eliminates a large number of redundant lstat calls and should give us lower latency and higher throughput when processing notifications from the kernel.

That benchmark changes the 2 seconds it takes for 260k lstat calls for todays implementation to a fraction of a second with this more optimal version.

In addition to the individual test plans in these diffs, I run through the mercurial test suite with the tip of this branch:

```
 WATCHMAN_SOCK=/opt/facebook/watchman/var/run/watchman/wez-state/sock ./run-tests.py --blacklist=blacklists/fsmonitor --extra-config="extensions.fsmonitor=" -j 3
```

and observed no new failures vs. the currently deployed watchman.

### DIFF
--- a/pending.c
+++ b/pending.c
@@ -10,6 +10,54 @@ static const struct flag_map kflags[] = {
   {0, NULL},
 };
 
+// Since the tree has no internal knowledge about path structures, when we
+// search for "foo/bar" it may return a prefix match for an existing node
+// with the key "foo/bard".  We use this function to test whether the string
+// exactly matches the input ("foo/bar") or whether it has a slash as the next
+// character after the common prefix ("foo/bar/" as a prefix).
+static bool is_path_prefix(const char *path, size_t path_len,
+                           size_t common_prefix) {
+  if (common_prefix == path_len) {
+    return true;
+  }
+
+  if (common_prefix > path_len) {
+    return false;
+  }
+
+  return path[common_prefix] == WATCHMAN_DIR_SEP
+#ifdef _WIN32
+         // Windows allows both kinds of slashes
+         || path[common_prefix] == '/'
+#endif
+      ;
+}
+
+// Helper to un-doubly-link a pending item.
+static inline void unlink_item(struct watchman_pending_collection *coll,
+                               struct watchman_pending_fs *p) {
+  if (coll->pending == p) {
+    coll->pending = p->next;
+  }
+  if (p->prev) {
+    p->prev->next = p->next;
+  }
+  if (p->next) {
+    p->next->prev = p->prev;
+  }
+}
+
+// Helper to doubly-link a pending item to the head of a collection.
+static inline void link_head(struct watchman_pending_collection *coll,
+                             struct watchman_pending_fs *p) {
+  p->prev = NULL;
+  p->next = coll->pending;
+  if (coll->pending) {
+    coll->pending->prev = p;
+  }
+  coll->pending = p;
+}
+
 /* Free a pending_fs node */
 void w_pending_fs_free(struct watchman_pending_fs *p) {
   w_string_delref(p->path);
@@ -20,23 +68,20 @@ void w_pending_fs_free(struct watchman_pending_fs *p) {
 bool w_pending_coll_init(struct watchman_pending_collection *coll) {
   coll->pending = NULL;
   coll->pinged = false;
-  coll->pending_uniq = w_ht_new(WATCHMAN_BATCH_LIMIT, &w_ht_string_funcs);
-  if (!coll->pending_uniq) {
-    return false;
-  }
   if (pthread_mutex_init(&coll->lock, NULL)) {
     return false;
   }
   if (pthread_cond_init(&coll->cond, NULL)) {
     return false;
   }
+  art_tree_init(&coll->tree);
   return true;
 }
 
 /* destroy a pending_coll */
 void w_pending_coll_destroy(struct watchman_pending_collection *coll) {
   w_pending_coll_drain(coll);
-  w_ht_free(coll->pending_uniq);
+  art_tree_destroy(&coll->tree);
   pthread_mutex_destroy(&coll->lock);
   pthread_cond_destroy(&coll->cond);
 }
@@ -49,7 +94,8 @@ void w_pending_coll_drain(struct watchman_pending_collection *coll) {
     w_pending_fs_free(p);
   }
 
-  w_ht_free_entries(coll->pending_uniq);
+  art_tree_destroy(&coll->tree);
+  art_tree_init(&coll->tree);
 }
 
 /* compute a deadline on entry, then obtain the collection lock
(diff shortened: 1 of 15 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @Vaelatern and @anirbanr-fb to be potential reviewers.

facebook-github-bot-5:

facebook-github-bot-3:

facebook-github-bot-4:

facebook-github-bot-9:

facebook-github-bot-1:

facebook-github-bot-3:

thirdparty/libart/src/art.c
```diff
@@ -21,6 +33,44 @@ static uint32_t __inline __builtin_ctz(uint32_t x) {
 }
 #endif
 
+// The ART implementation requires that no key be a full prefix of an existing
+// tree during insertion.  In practice this means that each key must have a
```
sid0: `s/tree/key`

thirdparty/libart/src/art.c
```diff
@@ -215,15 +265,15 @@ static int check_prefix(const art_node *n, const unsigned char *key, int key_len
 
 /**
  * Checks if a leaf matches
- * @return 0 on success.
+ * @return true if the key is an exact match.
  */
-static int leaf_matches(const art_leaf *n, const unsigned char *key, int key_len, int depth) {
-    (void)depth;
+static bool leaf_matches(const art_leaf *n, const unsigned char *key, int key_len) {
```
sid0: We've effectively forked libart at this point, then :(

thirdparty/libart/src/art.c
```diff
@@ -21,6 +33,44 @@ static uint32_t __inline __builtin_ctz(uint32_t x) {
 }
 #endif
 
+// The ART implementation requires that no key be a full prefix of an existing
+// tree during insertion.  In practice this means that each key must have a
+// terminator character.  One approach is to ensure that the key and key_len
+// includes a physical trailing NUL terminator when inserting C-strings.
+// This doesn't help a great deal when working with binary strings that may be
+// a slice in the middle of a buffer that has no termination.
```
sid0: Ah, so you basically want to avoid making a copy of each binary string then?

thirdparty/libart/src/art.c
```diff
@@ -21,6 +33,44 @@ static uint32_t __inline __builtin_ctz(uint32_t x) {
...
+//
+// To facilitate this the key_at() function is used to look up the byte
+// value at a given index.  If that index is 1 byte after the end of the
+// key, we synthesize a fake NUL terminator byte.
```
sid0: I'd add a note here that this still assumes that the binary strings do not have embedded null bytes (which is valid because we only insert paths, and paths can never contain null bytes). If the binary strings did have embedded NULs then this approach wouldn't work.

thirdparty/libart/src/art.c
```diff
@@ -21,6 +33,44 @@ static uint32_t __inline __builtin_ctz(uint32_t x) {
...
+// @return the value of the key at the supplied index.
+static inline unsigned char key_at(const unsigned char *key, int key_len, int idx) {
+    if (idx == key_len) {
+        // Implicit terminator
+        return 0;
+    }
+#if !ART_SANITIZE_ADDRESS
+    // If we were built with -fsanitize=address, let ASAN catch this,
+    // otherwise, make sure we blow up if the input depth is out of bounds.
+    w_assert(idx >= 0 && idx <= key_len,
```
sid0: Does this assert work in release builds?

thirdparty/libart/src/art.c
```diff
@@ -241,10 +291,10 @@ void* art_search(const art_tree *t, const unsigned char *key, int key_len) {
     while (n) {
         // Might be a leaf
         if (IS_LEAF(n)) {
-            n = (art_node*)LEAF_RAW(n);
+            art_leaf *leaf = LEAF_RAW(n);
```
sid0: Are these changes related to this diff? Having these included is making this harder to read.

watchman_log.h
```diff
@@ -0,0 +1,47 @@
...
+void w_log(int level, WATCHMAN_FMT_STRING(const char *fmt), ...)
+#ifdef __GNUC__
+  __attribute__((format(printf, 2, 3)))
+#endif
+;
+
+// Similar to assert(), but uses W_LOG_FATAL to log the stack trace
+// before giving up the ghost
+#ifdef NDEBUG
+# define w_assert(e, ...) ((void)0)
```
sid0: Ah, guess it's not.

pending.c
```diff
@@ -98,14 +144,132 @@ void w_pending_coll_unlock(struct watchman_pending_collection *coll) {
...
+  unused_parameter(value);
+
+  if ((p->flags & W_PENDING_CRAWL_ONLY) == 0 && key_len > ctx->root->len &&
+      is_path_prefix((const char *)key, key_len, ctx->root->len) &&
+      !is_possibly_a_cookie(p->path)) {
+    // Unlink the child from the pending index.
+    unlink_item(ctx->coll, p);
+    // and completely free it.
+
+#if 0
```
sid0: Make this a debug call or remove it?

sid0:
`use art to optimize pending list construction` looks fine overall, but shouldn't it be after the artificial-prefix change rather than before? Seems like we're going to run into the prefix issue pretty quickly with the pending list.

pending.c
```diff
@@ -169,21 +339,28 @@ void w_pending_coll_append(struct watchman_pending_collection *target,
...
     }
 
-    p->next = target->pending;
-    target->pending = p;
-    w_ht_set(target->pending_uniq, w_ht_ptr_val(p->path), w_ht_ptr_val(p));
+    if (is_obsoleted_by_containing_dir(target, p->path)) {
+      w_pending_fs_free(p);
+      continue;
+    }
+    prune_obsoleted_children(target, p->path, p->flags);
```
sid0: One general comment here is that it is probably possible to split this up so that `is_obsoleted_by_containing_dir` is in a separate diff from the `prune_obsoleted_children` function (both really seem to add on to each other). But it's ok :)

pending.c
```diff
@@ -98,14 +144,132 @@ void w_pending_coll_unlock(struct watchman_pending_collection *coll) {
...
+  }
+
+  return 0;
+}
+
+// if there are any entries that are obsoleted by a recursive insert,
+// walk over them now and mark them as ignored.
+static void prune_obsoleted_children(struct watchman_pending_collection *coll,
+                                     w_string_t *path, int flags) {
+  if ((flags & (W_PENDING_RECURSIVE | W_PENDING_CRAWL_ONLY)) ==
```
sid0: As a casual reader, I'd be surprised that there's a condition in here -- it might make more sense to call this `maybe_prune_obsoleted_children`

root.c
```diff
@@ -498,12 +498,12 @@ bool w_root_process_pending(w_root_t *root,
...
 
   w_log(W_LOG_DBG, "processing %d events in %s\n",
-      w_ht_size(coll->pending_uniq), root->root_path->buf);
+      w_pending_coll_size(coll), root->root_path->buf);
 
   // Steal the contents
   pending = coll->pending;
   coll->pending = NULL;
-  w_ht_free_entries(coll->pending_uniq);
+  w_pending_coll_drain(coll);
```
sid0: I don't understand this change -- could you explain it?

tests/pending_test.c
```diff
@@ -0,0 +1,144 @@
...
+
+    gettimeofday(&start, NULL);
+    for (item = list.avail - 1; item >= list.pending; item--) {
+      w_pending_coll_add(&coll, item->path, item->now, item->flags);
+    }
+
+    drained = process_items(&coll);
+
+    gettimeofday(&end, NULL);
+    diag("took %.3fs to reverse insert %u items into pending coll",
```
sid0: What does "reverse insert" mean here? Might be better to call this "top down" or "bottom up".

root.c
wez: This is fixing a mild layering violation.  This bit of code wants to steal the linked list of pending items, but wants to make sure that the pending list doesn't reference them any more.  Previously this was done by ensuring that the pending_uniq hash got emptied, but since that hash was removed by this diff it can't do that any more.

Really this code should have been calling `w_pending_coll_drain`; that function pops anything from the list (which will be empty in this case, because we just stole it) and then empties the hash.

After this diff, we use the `art` instead of the hash.  There isn't an equivalent function to `w_ht_free_entries` for `art`, instead we have to destroy and re-initialize; we could chose to continue the layering violation here, but it would look uglier :-)

thirdparty/libart/src/art.c
wez: Yeah, not so much at insertion time, but when we perform a prefix search we don't want to have to make a copy, especially if there are no matches for the prefix.

thirdparty/libart/src/art.c
wez: yeah, it didn't seem like the upstream was very open to making changes in any case, and I've got my eye on a further change to allow holding a reference to our `w_string_t` in the leaves rather than making a copy.

facebook-github-bot-0:

facebook-github-bot-8:

facebook-github-bot-9:

PR #260, opened 2016-06-06
Added a unicode filename to test_subscribe integration test.

dhruvsinghal:
Tested it under Python 2.7.5 and 3.5.1.

### DIFF
--- a/tests/integration/WatchmanTestCase.py
+++ b/tests/integration/WatchmanTestCase.py
@@ -29,6 +29,15 @@ else:
 def norm_path(name):
     return os.path.normcase(os.path.normpath(name))
 
+# TODO: This normalization will not be needed once we have full unicode support
+# in place as per
+# https://github.com/facebook/watchman/wiki/Better-Unicode-handling-plan
+def conv_path_to_bytes(path):
+    if isinstance(path, pywatchman.compat.UNICODE):
+        return path.encode('utf8')
+    else:
+        return path
+
 class WatchmanTestCase(unittest.TestCase):
 
     def requiresPersistentSession(self):
@@ -188,7 +197,6 @@ class WatchmanTestCase(unittest.TestCase):
         files = self.normWatchmanFileList(res['files'])
         self.last_file_list = files
         return files
-
     def waitForSync(self, root):
         """ ensure that watchman has observed any pending file changes
             This is most useful after mutating the filesystem and before
@@ -208,6 +216,16 @@ class WatchmanTestCase(unittest.TestCase):
     def normFileList(self, files):
         return sorted(map(norm_path, files))
 
+    def assertFileListsEqual(self, list1, list2, message=None):
+        list1 = [conv_path_to_bytes(f) for f in list1]
+        list2 = [conv_path_to_bytes(f) for f in list2]
+        self.assertEqual(list1, list2, message)
+
+    def fileListsEqual(self, list1, list2):
+        list1 = [conv_path_to_bytes(f) for f in list1]
+        list2 = [conv_path_to_bytes(f) for f in list2]
+        return list1 == list2
+
     # Wait for the file list to match the input set
     def assertFileList(self, root, files=[], cursor=None,
                        relativeRoot=None, message=None):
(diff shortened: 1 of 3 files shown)


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @sid0 and @wez to be potential reviewers.

facebook-github-bot-8:

sid0:
cc @c-jiph

facebook-github-bot-0:

tests/integration/WatchmanTestCase.py
```diff
@@ -29,6 +29,13 @@
 def norm_path(name):
     return os.path.normcase(os.path.normpath(name))
 
+def conv_path_to_bytes(path):
+    if (not pywatchman.compat.PYTHON3 and isinstance(path, unicode)) or \
```
sid0: Could you add a TODO comment here saying that this will go away/no longer need to be normalized once https://github.com/facebook/watchman/wiki/Better-Unicode-handling-plan is done?

sid0: Also, this would probably be better expressed by just adding `UNICODE` to `pywatchman.compat` or similar.

sid0: (`UNICODE` would be `unicode` on Python 2 and `str` on Python 3.)

tests/integration/WatchmanTestCase.py
```diff
@@ -208,6 +214,16 @@ def normWatchmanFileList(self, files):
     def normFileList(self, files):
         return sorted(map(norm_path, files))
 
+    def assertFileListsEqual(self, list1, list2, message=None):
+        list1 = list(map(conv_path_to_bytes, list1))
```
sid0: This is more easily expressed as:

```
list1 = [conv_path_to_bytes(f) for f in list1]
```

tests/integration/test_subscribe.py
```diff
@@ -200,15 +200,18 @@ def test_immediate_subscribe(self):
                               accept=lambda x: self.wlockExists(x, False))
         self.assertNotEqual(None, dat)
 
+    @unittest.skipIf(os.name == 'nt', 'win')
```
sid0: You shouldn't skip the whole test here because most of it actually works on Windows. You should either drop the Unicode name on Windows or create a separate test (which can be simplified from the original `test_subscribe`.

sid0:
Thanks for doing this! Looks really good except for a few comments. I'll be happy to land this once you're done :)

facebook-github-bot-2:

dhruvsinghal:
@sid0 I made the changes.

sid0:
Could you combine both your changes into one commit? That'll keep it logical.

tests/integration/test_subscribe.py
```diff
@@ -251,7 +251,64 @@ def matchesRecrawledDir(subdata):
...
+        for item in dat:
+            if 'warning' in item:
+                warn = self.decodeBSERUTF8(item['warning'])
+                break
+        self.assertRegexpMatches(warn, r'Recrawled this watch')
+
+    # TODO: Assimilate this test into test_subscribe when Watchman gets
+    # unicode support.
+    # TODO: Correctly test subscribe with unicode on Windows.
+    @unittest.skipIf(os.name == 'nt', 'win')
```
sid0: This doesn't exist on Python 2.6, so you'll need to import `unittest2` there instead. Check out how it's done in `WatchmanTestCase.py`.

sid0:
Other than those two changes, looks good! Looks like you may also need to rebase and maybe locally resolve conflicts.

facebook-github-bot-7:

facebook-github-bot-5:

tests/integration/test_subscribe.py
```diff
@@ -251,7 +251,65 @@ def matchesRecrawledDir(subdata):
...
+                warn = self.decodeBSERUTF8(item['warning'])
+                break
+        self.assertRegexpMatches(warn, r'Recrawled this watch')
+
+    # TODO: Assimilate this test into test_subscribe when Watchman gets
+    # unicode support.
+    # TODO: Correctly test subscribe with unicode on Windows.
+    def test_subscribe_unicode(self):
+        if os.name == 'nt':
+            self.skipTest('This currently does not work on Windows.')
```
sid0: No, this won't work either because the test skipping functionality just doesn't exist in Python 2.6's unittest. Just do what's done here: https://github.com/facebook/watchman/blob/master/tests/integration/WatchmanTestCase.py#L11

sid0: (And switch back to the `skipIf` call because it's cleaner)

facebook-github-bot-6:

sid0:
Pushed as f03d41acf48f0c93e2c6ffae3096756cc2a821bf, thanks!

PR #262, opened 2016-06-07
migrate legacy trigger test programs to python

wez:
Summary: I spent a couple of hours tracking down what appeared to be
a regression on Linux yesterday, but is actually a regression in a local
development build of hhvm for scripts.

We're overdue to move some of these legacy tests to our new python test
suite anyway, so this ports the helper programs that are registered as
triggers to use python instead of php.

Test Plan: `./runtests.py tests/integration/trigger.php` exercises this.

Prior to this change, the affected hhvm build would delay the start of
the trigger programs by several seconds (presumably some file locking
contention while running the compiler) and cause this test to take about
15 seconds.

After this change its runtime is back down to less than 2 seconds.

Reviewers: sid0

Subscribers:

Tasks:

Blame Revision:


facebook-github-bot-9:

wez:
Pushed as a920fe5290c4a31e1894f63d6a7dbfb843c0663a

PR #263, opened 2016-06-07
remove _ignore_watchmanconfig option

wez:
The foul purposes for which it was needed are no more.  Remove it.

Test Plan: make integration


wez:
Pushed this as d8aafa1fd44b1a0f15473a7de1ff87c94e5d9a76

PR #264, opened 2016-06-08
disable use of bulkstat on older Darwin kernels

wez:
Summary: I've seen this problem once in the past couple of months
and we worked around it by setting the kill switch option.

I remembered this just now while prepping for the 4.6 release; I don't
want to break the user that worked around this (the update will blow
away their modified watchman.json file), so let's tuck in this check
to set the default.

Test Plan: I added a W_LOG_FATAL to log the values of `safe`, `maj`,
`min`, `patch` and visually verified that they match my system and that
`safe` is `true`, indicating that we are safe to use bulk stat.

### DIFF
--- a/opendir.c
+++ b/opendir.c
@@ -3,6 +3,7 @@
 
 #include "watchman.h"
 #ifdef __APPLE__
+# include <sys/utsname.h>
 # include <sys/attr.h>
 # include <sys/vnode.h>
 #endif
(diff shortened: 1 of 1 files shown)


opendir.c
```diff
@@ -384,6 +385,32 @@ static DIR *opendir_nofollow(const char *path)
 #endif
 }
 
+#ifdef HAVE_GETATTRLISTBULK
+// I've seen bulkstat report incorrect sizes on kernel version 14.5.0.
```
sid0: Could you document what version of OS X this is?

wez: will do

wez:
Pushed as ef2e6d6f88736f6726865d26aa71a6572f8672cf

PR #265, opened 2016-06-08
listener: allow writing to socket for entities with access

sid0:
Without this, on Linux you can't connect to the socket.


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @wez, @Vaelatern and @anirbanr-fb to be potential reviewers.

PR #266, opened 2016-06-09
avoid a race condition during crawling

wez:
Summary: I suspect that the issue observed by @simpkins was due to a
race during the initial crawl.

The explanation of the race is inline in comments in the code.

Test Plan: I don't think there is a reliable way to trigger this and/or prove it short
of rolling it out.  The integration tests continue to pass as they did
before.


sid0:
lgtm

wez:
Pushed as cbbb8b50e5977a10c8510e3965597aa5ac9569f2

PR #267, opened 2016-06-09
Add ASAN support to configure

sid0:
This allows configuring GCC or clang with ASAN.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,6 +95,10 @@ if test "$WATCHMAN_STATE_DIR" != "no"; then
 fi
 AM_CONDITIONAL(WATCHMAN_STATE_DIR, [test "$WATCHMAN_STATE_DIR" != "no"])
 
+AC_ARG_ENABLE(asan, [
+  --enable-asan           Enable address sanitizer options
+])
+
 want_python=yes
 AC_ARG_WITH(python, [
   --with-python=PATH      Enable Python bindings. PATH is location of python.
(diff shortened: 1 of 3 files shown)


facebook-github-bot-9:
By analyzing the blame information on this pull request, we identified @wez, @Vaelatern and @anirbanr-fb to be potential reviewers.

facebook-github-bot-2:
@sid0 updated the pull request.

configure.ac
```diff
@@ -268,6 +275,9 @@ fi
 if test -n "$GCC" ; then
   CFLAGS="$CFLAGS -Wall -Wextra -Wdeclaration-after-statement -g -gdwarf-2 -fno-omit-frame-pointer"
 fi
+if test -n "$GCC" -a "$enable_asan" = "yes" ; then
+  CFLAGS="$CFLAGS -fsanitize=address"
```
wez: I normally add these flags to LDFLAGS too

watchman.h
```diff
@@ -100,6 +100,18 @@ extern char **environ;
 # define PRIsize_t "zu"
 #endif
 
+#if defined(__clang__)
+# if __has_feature(address_sanitizer)
+#  define ASAN_ENABLED 1
```
wez: missing WATCHMAN prefix here

facebook-github-bot-5:
@sid0 updated the pull request.

wez:
FWIW, I typically want to set -O0 when using ASAN so that inlined functions don't get lost.  Not sure if we want to force that together with this configure option or add another one just for turning off optimization... I'm actually not sure if we already have one of those or not :-)

sid0:
Hmm, good point re inlined functions. I didn't find anything related to setting `-O` options -- I think a configure option makes sense. What do you want it to be called?

wez:
maybe --disable-optimization?

sid0:
I'll call it `--disable-opt`

facebook-github-bot-9:
@sid0 updated the pull request.

wez:
optional/maybe for a later PR: we used to have an extra column in the test matrix on travis for valgrind enabled test runs, would be interesting to try that again but with ASAN and see if the results come back in a reasonable time.

Assuming the CI is all green for this though, I think this looks good to push as-is; thanks!

sid0:
Hmm, looks like the Travis OSX environment broke. OK I'll try tomorrow.

Yeah, I had planned to add this to the CI matrix.

facebook-github-bot-7:
@sid0 updated the pull request.

facebook-github-bot-4:
@sid0 updated the pull request.

sid0:
Pushed this series.

PR #271, opened 2016-06-09
travis: enable ASAN in CI

sid0:


sid0:
Looks like the builds now pass ASAN checks. Yay!

wez:
Looks good, thanks!

sid0:
Pushed (in the wrong order by accident, but that's ok)

PR #274, opened 2016-06-11
fixup prefix iteration

wez:
Summary: This was a bit of a surprise and caused some incorrect
behavior.  I added the logging first to gain some insight and found
that we were overly aggressive at pruning in delete_kids.

I tracked this down to the partial prefix matching code in the prefix
iterator; one of the cases is that the minimum node is located and used
as the basis for a vanilla recursive iteration.

So for example, we would be searching for `/foo/bar/baz` prefixes and
the tree would use `/foo/bar` as the root of a recursive iteration and
emit all the leaves under `/foo/bar` rather than just those that were
logically under `/foo/bar/baz`.

This condition triggers in `tests/integration/age.php` where we stress
test the node aging code and repeatedly create and remove nodes with
similar and same names.   The `w_assert` in the iterator callback
in `pending.c` triggers in non-prod builds in this case.

To address this, the prefix iteration needs to shim in a callback to
pre-check the leaves before they are passed to the user-supplied
callback, and that is what the bulk of the functional changes do in
this diff.

Test Plan: `make integration` would fail once I added the assert but
before I added the iterator callback shim.  After adding the shim,
all tests pass.

Capture the debug logs for updating a big repo via:

```
watchman --server-encoding=json -p log-level debug | egrep
'(delete_kid|maybe_prune|SKIP)' > /data/users/wez/blarg.log
```

and inspect and sanity check the output.

Reviewers: sid0

Subscribers:

Tasks: #11693500

Blame Revision:

### DIFF
--- a/pending.c
+++ b/pending.c
@@ -15,14 +15,18 @@ static const struct flag_map kflags[] = {
 // with the key "foo/bard".  We use this function to test whether the string
 // exactly matches the input ("foo/bar") or whether it has a slash as the next
 // character after the common prefix ("foo/bar/" as a prefix).
-static bool is_path_prefix(const char *path, size_t path_len,
+static bool is_path_prefix(const char *path, size_t path_len, const char *other,
                            size_t common_prefix) {
-  if (common_prefix == path_len) {
-    return true;
-  }
-
   if (common_prefix > path_len) {
     return false;
+  }
+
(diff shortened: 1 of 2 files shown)


sid0:
That's a really subtle bug. The change looks good to me.

wez:
Pushed as f0029886e6017aef824e7ad8de1b4a1eb851077c

PR #275, opened 2016-06-13
speculative fix for some fishy state-leave behavior

wez:
Summary: This has been hard to nail down, but we've seen some weird
crashes every so often where we see events like this in the log:

```
2016-06-10T21:34:06,035: [client:stm=0x63c5c0] implicitly vacating state hg.update on /data/users/wez/FOO due to client disconnect
2016-06-10T21:34:24,367: [client:stm=0x63c5c0] send_error_response: ["state-leave", "/data/users/wez/FOO", {"name": "hg.update", "metadata": {"rev": "hash1 hash2", "distance": 123, "status": "ok", "partial": false}}] failed: state hg.update is not asserted
```

this doesn't make sense because the the first line happens 20 seconds
before the second, but can't possibly run before the latter.

I haven't found a way to reproduce this even intermittently, let alone
reliably, so this diff adds some belt and suspenders stuff around the
cleanup of this so that we can hopefully get a better read on this
if/when it happens in the future.

Test Plan: make integration + CI

### DIFF
--- a/cmds/state.c
+++ b/cmds/state.c
@@ -209,32 +209,34 @@ static void leave_state(struct watchman_user_client *client,
     struct watchman_user_client *subclient = w_ht_val_ptr(iter.value);
     w_ht_iter_t citer;
 
-    if (w_ht_first(subclient->subscriptions, &citer)) do {
-      struct watchman_client_subscription *sub = w_ht_val_ptr(citer.value);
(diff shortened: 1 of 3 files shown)


facebook-github-bot-5:

sid0:
Sure, lgtm

PR #279, opened 2016-06-15
stave off a potential zombie horde

wez:
Summary: we had an incident where the new perf command was triggering
too frequently for some of our users and they wound up with a full
process table consisting of zombies.

This code is supposed to reap all the children, but it was somehow
not effective.  Staring at this a bit, it seems possible that the
waitpid() call may have gotten EINTR for some other signal that
was not in our signal set, and that didn't cause us to terminate.

In this situation, I think we may end up not realizing that we have
some children to reap and block for an extended period.

I've restructured the loop so that we can only suspend execution
if we got an ECHILD condition from waitpid, which should only
happen when we have waited for all eligible children.

Test Plan: make integration continues to succeed.

The nature of this issue is difficult to reproduce :-/

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -550,8 +550,30 @@ static void *child_reaper(void *arg)
   }
 #else
   while (!w_is_stopping()) {
+    int err;
+
+    // Poll for any finished child processes.
     w_reap_children(false);
-    sigsuspend(&sigset);
+    err = errno;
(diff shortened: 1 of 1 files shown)


sid0:
Yeah, this seems to be the least implausible explanation. lgtm

PR #280, opened 2016-06-16
add site-specific spawn command

wez:
Summary: in some situations, the system administrator may desire to have
    control over the context in which the watchman service is started.

    For example, if may be desirable to force the service to spawn a
    system service that was automatically created for the current user
    rather than just having watchman fork off a child.

    For this situation we now provide a `spawn_watchman_service`
    configuration key that is available in `/etc/watchman.json`.
    If this string is set, when the watchman CLI decides that it is
    appropriate to spawn the watchman service, rather than using the
    default spawning action, it will instead invoke `spawn_watchman_service`
    and forward all daemon appropriate watchman arguments to it.

    The value of `spawn_watchman_service` will be used as `argv[0]` when
    launching the spawner, so it must either be the full path to the
    spawning program, or be something that posix_spawn is able to resolve
    from the PATH of the current process.

    If the spawner returns with a non-zero exit status, the watchman command
    that invoked it will fail and report the problem to the user.

    The stdout and stderr of the spawner program are redirected to the
    watchman log file.

### DIFF
--- a/main.c
+++ b/main.c
@@ -932,6 +932,18 @@ const char *get_sock_name(void)
   return sock_name;
 }
 
+static void spawn_watchman(void) {
+#ifdef USE_GIMLI
+  spawn_via_gimli();
+#elif defined(__APPLE__)
+  spawn_via_launchd();
+#elif defined(_WIN32)
+  spawn_win32();
+#else
+  daemonize();
+#endif
+}
+
 int main(int argc, char **argv)
 {
   bool ran;
@@ -956,15 +968,7 @@ int main(int argc, char **argv)
         ran = try_client_mode_command(cmd, !no_pretty);
       }
     } else {
-#ifdef USE_GIMLI
-      spawn_via_gimli();
-#elif defined(__APPLE__)
-      spawn_via_launchd();
-#elif defined(_WIN32)
-      spawn_win32();
-#else
-      daemonize();
-#endif
+      spawn_watchman();
       ran = try_command(cmd, 10);
     }
   }
--- a/main.c
+++ b/main.c
@@ -34,10 +34,6 @@ static int json_input_arg = 0;
 #include <mach-o/dyld.h>
 #endif
 
-#if defined(USE_GIMLI) || defined(__APPLE__) || defined(_WIN32)
-# define SPAWN_VIA_LAUNCHER 1
-#endif
-
 static const char *compute_user_name(void);
 static void compute_file_name(char **strp, const char *user, const char *suffix,
                               const char *what);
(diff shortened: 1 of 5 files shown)


facebook-github-bot-9:

main.c
```diff
@@ -325,6 +320,61 @@ static void spawn_via_gimli(void)
...
+
+  if (WIFEXITED(res) && WEXITSTATUS(res) == 0) {
+    return;
+  }
+
+  if (WIFEXITED(res)) {
+    w_log(W_LOG_FATAL, "%s: exited with status %d\n", spawner,
+          WEXITSTATUS(res));
+  } else if (WIFSIGNALED(res)) {
+    w_log(W_LOG_FATAL, "%s: signalled with %d\n", spawner, WTERMSIG(res));
```
sid0: US english is "signaled"

tests/integration/test_site_spawn.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_spawner(self):
+      config = {
+          'spawn_watchman_service': os.path.join(os.path.dirname(__file__), 'site_spawn.py')
+      }
+
+      inst = WatchmanInstance.Instance(config=config)
+      stdout, stderr = inst.commandViaCLI(['version'])
+
+      parsed = json.loads(stdout)
```
sid0: Needs to be `stdout.decode('ascii')` to satisfy Python 3

tests/integration/test_site_spawn.py
```diff
@@ -0,0 +1,46 @@
...
+class TestSiteSpawn(unittest.TestCase):
+    def test_failingSpawner(self):
+      config = {
+          'spawn_watchman_service': os.path.join(os.path.dirname(__file__), 'site_spawn_fail.py')
+      }
+
+      inst = WatchmanInstance.Instance(config=config)
+      stdout, stderr = inst.commandViaCLI(['version'])
+
+      self.assertEqual('', stdout)
```
sid0: `self.assertEqual(b'', stdout)` to satisfy Python 3.

main.c
```diff
@@ -325,6 +320,61 @@ static void spawn_via_gimli(void)
 }
 #endif
 
+// Spawn watchman via a site-specific spawn helper program.
+// We'll pass along any daemon-appropriate arguments that
+// we noticed during argument parsing.
+static void spawn_site_specific(const char *spawner)
```
sid0: This whole function needs to be gated on Windows.

sid0:
Looks fine to me modulo comments. Thanks for doing this!

facebook-github-bot-4:

facebook-github-bot-4:

facebook-github-bot-3:

facebook-github-bot-4:

PR #282, opened 2016-06-17
take advantage of FSEventStreamSetExclusionPaths

wez:
Summary: this isn't indexed from the main fsevents docs, but it is a
public API; if you google for it, you will find it listed in changelogs.

The header file documents that it has a limit of 8 items.  From playing
around with this, if we exceed the limit, the function call will fail
with no explanation.

Because of this limit, it is advantageous to list the highest value
ignores in the first 8 entries of the `ignore_dirs` configuration
in `.watchmanconfig`, and we therefore need an ordered list of the
ignores; our current datastructures are unordered maps and that results
in a non-deterministic set of excluded paths on any given attempt to
watch the same dir.

Test Plan: `make integration` succeeds.  I also added some
instrumentation to see what the behavior is; the exact path listed in
the exclusion paths array is delivered to our stream, but any child
paths are ignored.  That's good enough for us.

### DIFF
--- a/ignore.c
+++ b/ignore.c
@@ -11,6 +11,7 @@
 
 bool w_ignore_init(struct watchman_ignore *ignore) {
   ignore->ignore_vcs = w_ht_new(2, &w_ht_string_funcs);
+  ignore->dirs_vec = NULL;
   if (!ignore->ignore_vcs) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


facebook-github-bot-9:

sid0:
Needs some tests, I guess?

wez:
`tests/integration/ignore.php` continues to pass (there's a more involved test at the bottom).  I don't really want to add instrumentation into the dispatcher to make a stronger assertion in the test harness as we're already sensitive to perf in there :-/

sid0:
Ah, nvm, I saw that that config option defaults to `true`.

What's the earliest version of OS X this is present in? Do we need to document this somewhere?

sid0:
(Looks fine otherwise.)

wez:
10.9 (Mavericks) and up according to the header file.  Hopefully we don't have anybody on anything lower...

wez:
that would bump our requirement up from 10.7 to 10.9 starting in this version.  I think I'm OK with that; I'm pretty sure our direct set of machines are all on something much newer than 10.9 in any case.

wez:
OK, modified this to add a configure check and ifdef so that we don't completely break this for people deploying to older versions (there's a DEPLOYMENT_TARGET env var that I think you can set to select this at compile time in the apple tool chain).

Pushed as ce411a5f3cf1f21fc67146f5874c487d60e48c8d

PR #283, opened 2016-06-17
poison when the system hits the number-of-files limits

wez:
If the watchman process itself runs out of files, or if the system
file table is full, we will have missed some notifications and have
no way to self-remediate, so let's poison ourselves.

There isn't a great way to test this in CI :-/


sid0:
LGTM

It'd be worth documenting this too

wez:
Added docs, also added watch-del-all to the list of poison-immune commands.
Pushed as 4a9e510

PR #285, opened 2016-06-17
Nominally added bser-v2 as a new capability.

dhruvsinghal:


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @sid0 and @wez to be potential reviewers.

python/pywatchman/capabilities.py
```diff
@@ -47,6 +47,7 @@ def parse_version(vstr):
     "term-dirname": "3.1",
     "term-idirname": "3.1",
     "wildmatch": "3.7",
+    "bser-v2": "4.6.0"
```
wez: I'd actually like this synthetic cap thing to go away; no older version of watchman supports this anyway, so it doesn't make too much sense to add new features in here.  The main purpose of this was to bridge the gap over the period where we added some of these new features but hadn't added capability support yet

dhruvsinghal: Done.

facebook-github-bot-9:

PR #286, opened 2016-06-19
eliminate a string copy in the fsevents callback

wez:
This removes a stack buffer and a copy.  We used to require that we have
a NUL terminated string for some flavor of the old ignore list
processing, but the new art based code doesn't have this restriction.

Optimize slightly by avoiding a lock/signal if we ignored everything
that was in the notification callback, and by deferring the logging
to the thread that consumes these.


sid0:
Looks reasonable

wez:
pushed as df33ffb409bd230184383feca24eef79c73ff6b7

PR #287, opened 2016-06-19
Resync from fsevents journal on UserDropped

wez:
This PR implements a new configurable feature that will attempt to resync from the persistent fsevents journal in the event of a user-space overflow.

In my recent performance tests I've been unable to trigger a UserDropped event while watchman is running under instrumentation.  Furthermore, the performance data shows that the processing callback takes only a handful of milliseconds to run, with the critical section taking microseconds.  This leads me to believe that the root cause of the UserDropped events is not tardiness in the processing loop, but rather that we are not getting the CPU in a timely fashion.

Therefore it seems reasonable that we should be able to successfully resync from the fsevents journal when we do eventually get back on the CPU.

The approach taken here is to refactor the code that creates the stream such that we can create more than one, and pass in the event id as the since parameter.

Initially we sync from the current point in time, but if we do experience a UserDropped event we will set up a new stream using the last observed good event id.

This resync is not as heavyweight as a full recrawl because fsevents promises to tell us about the changes that we didn't see; we're effectively rewinding the stream to resync.  This should cut out a lot of recursive crawling IO and take less time over all.

Testing: I've spent a couple of hours and have been unable to trigger a UserDropped event with the 4.6 release candidate code thus far (my test case for this involves changing the sparse status of a deep tree), but our telemetry shows that this has happened for a couple of our mac users.  I'm going to try some build intensive tests when I get back into the office.

### DIFF
--- a/tests/integration/test_fsevents_resync.py
+++ b/tests/integration/test_fsevents_resync.py
@@ -1,0 +1,57 @@
+# vim:ts=4:sw=4:et:
+# Copyright 2016-present Facebook, Inc.
+# Licensed under the Apache License, Version 2.0
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+# no unicode literals
+
+import WatchmanTestCase
+import json
+import tempfile
+import os
+import os.path
+import sys
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+@unittest.skipIf(sys.platform != 'darwin', 'N/A unless macOS')
+class TestFSEventsResync(WatchmanTestCase.WatchmanTestCase):
+
+    def test_resync(self):
+        root = self.mkdtemp()
+        with open(os.path.join(root, '.watchmanconfig'), 'w') as f:
+            f.write(json.dumps({
+                'fsevents_try_resync': True
+            }))
+
+        self.watchmanCommand('watch', root)
+        self.touchRelative(root, '111')
+        self.assertFileList(root, ['.watchmanconfig', '111'])
+
+        res = self.watchmanCommand('query', root, {
+            'fields': ['name']})
+        self.assertTrue(res['is_fresh_instance'])
+        clock = res['clock']
+
+        dropinfo = self.watchmanCommand('debug-fsevents-inject-drop', root)
+        self.assertTrue('last_good' in dropinfo, dropinfo)
+
+        # We expect to see the results of these two filesystem operations
+        # on our next query, and not see evidence of a recrawl
+        os.unlink(os.path.join(root, '111'))
+        self.touchRelative(root, '222')
+
+        res = self.watchmanCommand('query', root, {
+            'since': clock,
+            'expression': ['exists'],
+            'fields': ['name']})
+        self.assertFalse(res['is_fresh_instance'], res)
+        self.assertTrue('warning' not in res, res)
+        self.assertEqual(self.normWatchmanFileList(res['files']),
+                         self.normFileList(['222']))
+
--- a/watcher/fsevents.c
+++ b/watcher/fsevents.c
@@ -21,6 +21,7 @@ struct fse_stream {
   FSEventStreamEventId last_good;
   bool lost_sync;
   bool inject_drop;
+  CFUUIDRef uuid;
 };
 
 struct fsevents_root_state {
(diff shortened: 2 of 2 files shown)


facebook-github-bot-9:

facebook-github-bot-9:

facebook-github-bot-9:

watcher/fsevents.c
```diff
@@ -123,62 +129,45 @@ static void fse_pipe_callback(CFFileDescriptorRef fdref,
...
   double latency;
+  struct fse_stream *fse_stream = calloc(1, sizeof(*fse_stream));
 
-  w_set_thread_name("fsevents %.*s", root->root_path->len,
-      root->root_path->buf);
-
-  // Block until fsevents_root_start is waiting for our initialization
-  pthread_mutex_lock(&state->fse_mtx);
+  if (!fse_stream) {
+    root->failure_reason = w_string_new("OOM");
```
sid0: Presumably in an OOM you wouldn't be able to allocate a new string?

watcher/fsevents.c
```diff
@@ -123,62 +129,45 @@ static void fse_pipe_callback(CFFileDescriptorRef fdref,
   CFRunLoopStop(CFRunLoopGetCurrent());
 }
 
-static void *fsevents_thread(void *arg)
-{
-  w_root_t *root = arg;
+static void fse_stream_free(struct fse_stream *fse_stream) {
```
sid0: I think this should be `fse_stream_destroy` going by our standards.

sid0: (But I'm still not 100% clear about the distinction so I could be wrong)

watcher/fsevents.c
```diff
@@ -279,6 +280,9 @@ static void *fsevents_thread(void *arg)
   // Block until fsevents_root_start is waiting for our initialization
   pthread_mutex_lock(&state->fse_mtx);
 
+  state->attempt_resync_on_user_drop =
+      cfg_get_bool(root, "fsevents_try_resync", false);
```
sid0: Hmm.

* Is the plan here to switch the default to `true` at some point?
* Will there come a time when we can make this config option go away entirely?

watcher/fsevents.c
```diff
@@ -79,6 +80,12 @@ static void fse_callback(ConstFSEventStreamRef streamRef,
   unused_parameter(streamRef);
 
   if (!stream->lost_sync) {
+    // This is to facilitate testing via debug-fsevents-inject-drop
+    if (stream->inject_drop) {
+      stream->lost_sync = true;
+      goto do_resync;
```
sid0: Don't you need to reset `stream->inject_drop` to false?

tests/integration/test_fsevents_resync.py
```diff
@@ -0,0 +1,51 @@
...
+import os.path
+import sys
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+@unittest.skipIf(sys.platform != 'darwin', 'N/A unless macOS')
+class TestFSEventsResync(WatchmanTestCase.WatchmanTestCase):
```
sid0: Where are you setting the config option to allow resyncs?

watcher/fsevents.c
```diff
@@ -233,13 +237,48 @@ static struct fse_stream *fse_stream_make(w_root_t *root,
...
   }
   fse_stream->root = root;
 
+  if (stat(root->root_path->buf, &st)) {
+    *failure_reason = w_string_make_printf(
+        "failed to stat(%s): %s\n", root->root_path->buf, strerror(errno));
+    goto fail;
+  }
+
+  fse_stream->uuid = FSEventsCopyUUIDForDevice(st.st_dev);
```
sid0: Maybe add a comment referencing the man page here?

sid0:
OK, I'll probably have another pass over it.

facebook-github-bot-9:

facebook-github-bot-9:

watcher/fsevents.c
sid0: Not sure if I'm doing github right -- did you mean to keep this the same?

watcher/fsevents.c
```diff
@@ -81,11 +85,65 @@ static void fse_callback(ConstFSEventStreamRef streamRef,
...
+            // If we fail, then we allow the UserDropped event to propagate
+            // to the consumer thread which has existing logic to schedule
+            // a recrawl.
+            w_string_t *failure_reason = NULL;
+            struct fse_stream *replacement =
+                fse_stream_make(root, stream->last_good);
+
+            if (!replacement) {
+              w_log(W_LOG_ERR,
+                    "Failed to rebuild fsevent stream (%.*s) while trying to "
```
sid0: `fsevents`

sid0:
LGTM

PR #288, opened 2016-06-20
make HINT_NUM_DIRS configurable

wez:
I noticed that this was still a fixed #define in the header.
Turn it into a `.watchmanconfig` option with the same default value.

This allows us to better match the value for our various repos.

### DIFF
--- a/watchman.h
+++ b/watchman.h
@@ -235,6 +235,7 @@ struct watchman_string {
 /* small for testing, but should make this greater than the number of dirs we
  * have in our repos to avoid realloc */
 #define HINT_NUM_DIRS 128*1024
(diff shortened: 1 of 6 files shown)


facebook-github-bot-1:

facebook-github-bot-9:

watchman.h
```diff
@@ -235,6 +235,7 @@ struct watchman_string {
 /* small for testing, but should make this greater than the number of dirs we
  * have in our repos to avoid realloc */
```
sid0: This comment is probably no longer necessary

sid0:
lgtm

wez:
Fixed comment and pushed as 381a5da

PR #289, opened 2016-06-22
Fix a nit in perf sampling, add a num_walked to query metadata

wez:
Two patches in this PR:

1. Fix missing format specifier for the num_results in a query
2. Add a counter for the number of nodes that a query walks over

This stuff doesn't currently have test infra, which is not great.  I'll probably squeeze the first of these into 4.6 and push the second out to 4.7, along with some tests.

### DIFF
--- a/cmds/subscribe.c
+++ b/cmds/subscribe.c
@@ -7,16 +7,20 @@ static bool subscription_generator(
     w_query *query,
     w_root_t *root,
     struct w_query_ctx *ctx,
-    void *gendata)
+    void *gendata,
+    int64_t *num_walked)
 {
   struct watchman_file *f;
   struct watchman_client_subscription *sub = gendata;
+  int64_t n = 0;
+  bool result = true;
 
   w_log(W_LOG_DBG, "running subscription %s %p\n",
       sub->name->buf, sub);
 
   // Walk back in time until we hit the boundary
   for (f = root->latest_file; f; f = f->next) {
+    ++n;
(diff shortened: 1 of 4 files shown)


facebook-github-bot-4:

cmds/subscribe.c
```diff
@@ -31,10 +34,12 @@ static bool subscription_generator(
     }
 
     if (!w_query_process_file(query, ctx, f)) {
+      *num_walked = n;
```
sid0: People are definitely going to forget doing this if they add return paths in the future. Can you add:

* a `done` label
* a bool `ret` value which can be set to `false` or `true`
* switch these internal returns to `goto done`?

query/eval.c
```diff
@@ -309,56 +338,76 @@ static bool path_generator(
...
   bool generated = false;
+  int64_t n = 0;
+  int64_t total = 0;
 
   unused_parameter(gendata);
 
   // Time based query
   if (ctx->since.is_timestamp || !ctx->since.clock.is_fresh_instance) {
-    if (!time_generator(query, root, ctx)) {
+    n = 0;
```
sid0: This is unnecessary.

sid0: Ah, I guess you want to do it for consistency.

sid0:
Looks fine other than my comment. I don't feel very strongly about it but from my experience a `goto` + handling this at the end usually leads to fewer mistakes down the line.

facebook-github-bot-7:

PR #290, opened 2016-06-22
Take another stab at the zombie problem

wez:
Summary: we've seen some zombies hanging out under watchman.  Manually
running `pkill -CHLD <pid-of-reaper-thread>` woke up the reaper to
collect them, so that implies that we have another race.

From looking at the code a bit harder, the race condition is likely that
the threads that we spawn from the state file are started before we mask
off SIGCHLD.

Because posix and threads are awesome, this means that the kernel can
choose to deliver SIGCHLD to an arbitrary thread that hasn't explicitly
blocked it.  This results in an EINTR somewhere, but not in the child
getting reaped.

This diff masks off SIGCHLD early, and arranges for the reaper to start
up before we spawn any threads, just to be sure that we have something
hanging around and ready to reap any children that might be start by
triggers loaded from the state file.

### DIFF
--- a/main.c
+++ b/main.c
@@ -151,10 +151,26 @@ static void run_service(void)
         hostname);
   }
 
+#ifndef _WIN32
+  // Block SIGCHLD by default; we only want it to be delivered
+  // to the reaper thread and only when it is ready to reap.
+  // This MUST happen before we spawn any threads so that they
+  // can pick up our default blocked signal mask.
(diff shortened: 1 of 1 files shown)


facebook-github-bot-9:

wez:
Pushed as 049834c66634738eb40fd8678757a057524cf84d

PR #291, opened 2016-06-23
avoid blocking on write during shutdown

wez:
Summary: we've seen a couple of instances where watchman has one
or more client threads blocked in write while the listener thread
is waiting for threads to finish.

The pre-condition appears to be that we need to have enough data
buffered up prior to initiating shutdown (possibly in conjunction
with a tardy client on the other end of the socket) that the blocking
write blocks the thread for some period of time.  The precise
conditions are hard to gather from the data we've gotten so far.

From inspecting the code, there are a couple of things we can do
to reduce the chances of blocking forever:

If we've been waiting a bit, send a signal to shake the thread out of a
blocking syscall.  That should bubble up as an EINTR and terminate
whatever loop they were in, and then they should notice that we're
shutting down.  Why not send the signal immediately?  I'm slightly
worried that it might interrupt some of the normal tear down in the
client thread if we send it too soon after signalling it via
client->ping.

Test Plan: it doesn't break make integration (I was concerned that
we'd make the shutdown a bit more aggro, but it seems ok).

Manually: `watchan --foreground` in one window, `tail -f log` in another
window, and in a third window `watchman shutdown-server`.  The shutdown
command completes with a valid shutdown response before the server
shuts down.

### DIFF
--- a/listener.c
+++ b/listener.c
@@ -407,7 +407,6 @@ static int get_listener_socket(const char *path)
 static struct watchman_client *make_new_client(w_stm_t stm) {
   struct watchman_client *client;
   pthread_attr_t attr;
-  pthread_t thr;
 
   pthread_attr_init(&attr);
   pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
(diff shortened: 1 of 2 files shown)


facebook-github-bot-5:

wez:
None of these tests can reproduce the problem scenario, so this is all a bit speculative.
Also, why not simply make the write call timeout? We'd need to introduce a way to express that we want a timeout along with enabling blocking, and then we'd need to tune to find a reasonable default timeout so that we avoid breaking clients.  It's more work and potentially more error prone for the !shutdown cases.

wez:
Pushed as 5e7d4a0

PR #294, opened 2016-06-28
Fix typo in link for inotify_add_watch fail

mdamien:
The link given in the error doesn't go to the section of the page since it's used dashes instead of underscore.

### DIFF
--- a/website/_docs/troubleshooting.markdown
+++ b/website/_docs/troubleshooting.markdown
@@ -102,7 +102,7 @@ The default log location may be overridden by the `--logfile` [Server Option](
 /watchman/docs/cli-options.html#quick-note-on-default-locations) explains what
 we mean by `<STATEDIR>`, `<TMPDIR>`, `<USER>` and so on.
 
-## Poison: inotify_add_watch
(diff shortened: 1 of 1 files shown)


facebook-github-bot-9:

wez:
Thanks for looking at this; actually, the problem is that the documentation generation system changed and broke these links.  Since older deployed versions of watchman use dashes in these link it would be great if we could find a way to persuade the docs to go back to using dashes; that way everything would still work.

facebook-github-bot-7:

facebook-github-bot-4:
@mdamien updated the pull request.

facebook-github-bot-4:
@mdamien updated the pull request.

mdamien:
yeah, I was thinking about doing that but it was kind of ugly in doc, here my solution is to be able to have both link working.

Do you think it's useful if we keep both links but promote the new one instead (by changing the link in the code) ? For now I stay with a simple fix.

wez:
Sorry for the delay; while this looks slightly ugly, I think this is the right fix.  Thanks for doing this!
Pushed as 4f4c0447ed50837243ff0200c1b2747714b0ccd9
I will rebuild the website this weekend.

PR #297, opened 2016-07-05
Update nodejs documentation

danez:
Hello, I made some small improvements to the nodejs docs, hope this is okay. I think especially the Int64 example is good to have.

Use === when comparing subscription.
Do the check that compares the subscription before looping over files.
Use forEach instead of for.
Add example about Int64, which is nice to know that the numbers returned are not plain numbers.


wez:
Thanks!  Pushed as 053ec69d79e2500f85a08f44b4a362c92785fe52
I'll rebuild and publish the docs to the website sometime over this weekend

PR #306, opened 2016-07-19
pywatchman: fix for windows 7 and earlier not supporting GetOverlappedResultEx

otrempe:


facebook-github-bot-4:
By analyzing the blame information on this pull request, we identified @wez, @sid0 and @dhruvsinghal to be potential reviewers.

facebook-github-bot-4:

facebook-github-bot-0:

python/pywatchman/__init__.py
```diff
@@ -114,16 +126,128 @@ def __init__(self):
...
-                                      ctypes.POINTER(OVERLAPPED), LPDWORD,
-                                      wintypes.DWORD, wintypes.BOOL]
-    GetOverlappedResultEx.restype = wintypes.BOOL
+    # Windows 7 and earlier does not support GetOverlappedResultEx
+    # The alternative is to use GetOverlappedResult and wait
+    # for read or write operation to complete. This is
+    # done be using CreateEvent and WaitForSingleObject.
+    # Windows 8 is version 6.2 - see the Operating System Version
+    # page on MSDN
+    WINDOWS_8_AND_LATER = sys.getwindowsversion() >= (6, 2)
```
wez: It would be great to exercise both variations on this code on the CI system.  We have some infrastructure for this sort of thing.  Take a look in `tests/integration/WatchmanTestCase.py` for the `expandConfigurations` method.

For windows I think we want to introduce a dimension there (Something like `('local', 'windows-7-bser')`) that forces us to use the Windows 7 implementation for one set of running the tests.

python/pywatchman/__init__.py
```diff
@@ -114,16 +126,128 @@ def __init__(self):
...
+
+        WaitForSingleObject = ctypes.windll.kernel32.WaitForSingleObject
+        WaitForSingleObject.argtypes = [wintypes.HANDLE, wintypes.DWORD]
+        WaitForSingleObject.restype = wintypes.DWORD
+    else:
+        GetOverlappedResultEx = ctypes.windll.kernel32.GetOverlappedResultEx
+        GetOverlappedResultEx.argtypes = [wintypes.HANDLE,
+                                          ctypes.POINTER(OVERLAPPED), LPDWORD,
+                                          wintypes.DWORD, wintypes.BOOL]
+        GetOverlappedResultEx.restype = wintypes.BOOL
```
wez: the approach that we took for this in the C implementation was to provide our own implementation of this function for Windows 7:

https://github.com/facebook/watchman/blob/master/stream_win.c#L49-L78

it would be easier to maintain if we could use broadly the same approach in here; that way it would be easier to see how to apply bug fixes from one to another.

It should also help to reduce the amount of conditional code throughout this module.  Would you mind taking a look at phrasing this change in that way?

otrempe: Sure!

otrempe: @wez 

Inspired by the fix made into the watchman source code to support overlapped operations on Windows 7, we did the same into pywatchman.
Here is the code:
```
    def _GetOverlappedResultEx(self, pipe, olap, nbytes, millis, alertable ):
        if WINDOWS_8_AND_LATER:
            return GetOverlappedResultEx( pipe, olap, nbytes, millis, alertable )

        # else, we fallback to a custom implementation of GetOverlappedResultEx
        # using GetOverlappedResult. This is the exact same implementation
        # that can be found in the watchman source code (stream_win.c)
        # This way, maintenance should be simplified.
        while True:
            if GetOverlappedResult( pipe, olap, nbytes, False ):
                # Result is available
                return True

            starttick = GetTickCount64()
            sleepexreturncode = SleepEx(millis, alertable)
            if sleepexreturncode == WAIT_IO_COMPLETION:
                SetLastError(WAIT_IO_COMPLETION)
                return False
            endtick = GetTickCount64()

            if millis != INFINITE:
                deltatick = endtick - starttick
                durationmillis = millis - deltatick
                if durationmillis <= 0:
                    # Out of time (We suspect we could get rid of this check
                    # using the return of SleepEx. Documentation says that
                    # the method will return 0 if the timeout has been reached).
                    SetLastError(WAIT_TIMEOUT)
                    return False
```

By debugging, we noticed a few things which deserve discussion.

First:
In the following line of the watchman code:
`millis = (DWORD)((ULONGLONG)millis - (end - start));`
reusing millis to store the calculation result forces it to be unsigned.
Since the timeout detection logic is based on having a negative delta (millis <= 0), the timeout detection won't occur unless `(end - start) == millis`, which would be kind of lucky.
Using long long variable type would be more appropriate. For example:
```
long long timeoutvssleepduration = (long long )((ULONGLONG)millis - (end - start));
if (timeoutvssleepduration <= 0) {…}
```

Second:
We noticed that using `GetTickCount64` is not always working as expected. We noticed several calls to a "timeouting" `SleepEx()` being shorter than the timeout value.
E.g.1: getting `deltatick == 1199` when `millis == 1200`
E.g.2: getting `deltatick == 1490` when `millis == 1500`

This leads to a broken detection of a real timeout.
That would deserve a tolerance if the current strategy is kept, at least in pywatchman (we did not debug in the watchman code)

Third:
We ain’t sure if it’s a python-only behavior (probably, since you would have noticed it before with the watchman service), but we observed that the first read op following a write op timeouts according to the call to `SleepEx`. On the second execution of the `while` loop, the call to  `GetOverlappedResult` always succeeds.

Here are the results of 7 executions of `hg status` in debug mode (we get a similar behavior when running not attached to a debugger.):

```
- - - - - - - clock: write (32 bytes)
- F - T F F F clock: read (13 bytes)
- - - - - - - clock: read (48 bytes)
- - - - - - - query: write (246 bytes)
F F F - T F T query: read (13 bytes)
- - - - - - - query: read (80 bytes)

F = Call failed at the first iteration loop but succeed on the second.
T = Call timed out (The timeout detection worked as expected.
There should be no F, only T… so it make obvious that the strategy to detect a timeout needs to be more robust
```

Our guess to explain the fails is that the read operation completes between the return from `GetOverlappedResult()` and the call to `SleepEx()`. This way `GetOverlappedResult` returns 0 BUT `SleepEx` timeout because the `IOCompletion` callback occurred just before. This is likely to happen because the call to `GetTickCount64()` delays just enough the call to `SleepEx`.
Note that adding a 10ms sleep before the call to `GetOverlappedResult( pipe, olap, nbytes, False )` fixes the problem because it allows the IO operation to complete.


Questions:
Would it be possible to verify this behavior on your side with the watchman service?
We are not setup to make those tests on our side. Maybe the situation doesn’t occur because calling Windows API from native code is faster than calling it from python using ctypes.windll.kernel32.*
To help reproducing the problem, maybe you could add a sleep between the calls to `GetOverlappedResult` and `SleepEx` to simulate a longer processing time of `GetTickCount64`

Conclusion:
While we agree that the solution we provided brings a lot of conditional code, it seems safer this way because of the use of an Event. Maybe the best approach would be to drop our wrapper, but still implement the `GetOverlappedResultEx` for Windows7 using the Event. Then, this improvement could be replicated in the C implementation.

wez: Thanks for being so thorough!  The implementation in the service was a bit speculative, since I didn't have access to a system running Windows 7, so it is quite possible that it is faulty.

First: good catch.  We should fix this in the C code too.

Second: hmm, what if we changed the `if durationmillis <= 0:` condition to be just `if durationmillis < 0:`?  That way we'd do another iteration at the end.

Third: Interesting.  Does it have to be a 10ms sleep?  Does a 0ms sleep have the same positive effect?   This probably also affects the C code and I'd love to have a solution in both areas.

We're not really geared up for debugging Windows code here and I don't have a lot of bandwidth to get hands on... are you able to build and run watchman's tests for yourself?  It may just be as simple as running `make` and then running `make integration` (there's a `make.bat` script in the top level that drives this).  It should be pretty self-contained; I think the only mandatory dependency is PHP for some of the legacy tests.

If you think it is easier to manage this using an Event, and can find a way to get that working without adding a lot of conditional code, I'd be happy to see that land!

Issue #305, opened 2016-07-19
Ability to skip invalidating directories for `touched` children


stuhood:
With watchman `4.5.0` on OSX and a query like:
```
expression=[
  'allof',  # All of the below rules must be true to match.
  ['not', ['pcre', r'^\..*', 'wholename']]  # Exclude files in hidden dirs.
]
```
... `touch`ing an _existing_ file `b/1` contained in a directory `b/` will invalidate the directory in addition to the file. Presumably this is because a directory listing that includes access/modification times might have changed?

It would be great to be able to (even on a best-effort basis) include something in triggered events to indicate _how_ a directory has changed. Our system does not use access/modification times, and thus doesn't need to invalidate directory listings for them.

Thanks for building a great system!

wez:
I wrote up a pretty long explanation of why we don't offer this here: http://stackoverflow.com/a/35753594/149111

With that out of the way, let's focus on helping you with your use case.  Can you tell me a bit more about it?  What I thought you were trying to do based on the description in this issue doesn't quite match up to the way I read the description in the pants issue you linked; I'm almost certainly missing something important :-)

Can you give some more detail on the full query that you're passing to watchman?  I'm particularly interested in how you work with the *since* generator and the *clock* values that watchman passes back.  In theory we can add a *since expression term* (**not** the same thing as the *since generator*!) to also constrain the results by *ctime*.  I think the trick will be in phrasing that filter without you taking watchman implementation details into your application.

stuhood:
Thanks for the quick response!

> doesn't quite match up to the way I read the description in the pants issue you linked

Yea, the linked issue is where I discovered this issue, rather than an expansion of this issue.

> Can you give some more detail on the full query that you're passing to watchman? 
The fully query (before I began to attempt to remove the filtering of directories) [is here](https://github.com/pantsbuild/pants/blob/master/src/python/pants/pantsd/service/fs_event_service.py#L62-L85), but I think that the minimization I included here is pretty straightforward.

The usecase is: we never use file timestamps. And so we'd like to avoid invalidating the parent directory when you touch an _existing_ file (the "existing" bit was missing from the original description: have added it now).

wez:
Can you phrase that in terms of what information you want from watchman and in which circumstances?  invalidation sounds like it is specific to some pants-internal data structure that I'm not familiar with.  The linked issue sounds like you may be computing a directory listing using a glob and that you recompute based on a change notification from watchman; is my interpretation accurate?

[...]

PR #307, opened 2016-07-19
avoid spuriously updating directories on darwin

wez:
Summary: the pantsbuild folks reported that they receive spurious
notifications for a parent directory when touching a child directory.

In tracing through the code, this is due to the logic in this diff
that desires to trigger a stat on the parent directory.  This behavior
was introduced in c6dbda20726d58f578bbca290c6a739c4e76d751 to ensure
that we notice mtime updates for the parent.

In this situation, since the event that triggered for the child node
had VIA_NOTIFY set, we will propagate this to the stat_path call when
looking at the parent and this will cause it to unconditionally mark
the directory node as changed (because the criteria is `via_notify ||
did_file_change()`).

I think the solution here is just to zero out the flags that we pass
down and that is what I've done here in this diff.  The existing
twodeep.php integration test continues to pass for me on the mac so this
doesn't obviously break things.  The diff that introduced this behavior
was Linux specific though, so I'm interested to see what our Linux CI
makes of this.

Refs: https://github.com/facebook/watchman/issues/305


facebook-github-bot-8:

sid0:
I agree that this makes sense, but please make the commit message a comment otherwise this will regress in the future.

PR #295, opened 2016-06-30
Adding the Java library, wrapping around the Watchman socket

cotizo:
This PR provides a Java library built by buck, which spawns the Watchman process and then interacts with it via the Watchman socket.

PR #309, opened 2016-07-20
Adding a java library around the Watchman socket

cotizo:
This PR provides a Java library built by buck, which spawns the Watchman process and then interacts with it via the Watchman socket.

This is a squashed down version of https://github.com/facebook/watchman/pull/295


wez:
Tweaked some whitespace and the ignore files slightly and pushed as f3e69138fd7af21b77afc33eb44576c37e64fcc9
Thanks for doing this!

PR #310, opened 2016-07-20
[wildmatch] Consume any number of consecutive slashes in pattern

bhamiltoncx:
This is to match Python's built-in `glob()` behavior.

Previously, globbing a pattern like `a//b/c` would not match the
path `a/b/c`. Since Unix treats mutiple slashes the same as one
slash, wildmatch should too.

### DIFF
--- a/query/match.c
+++ b/query/match.c
@@ -13,6 +13,7 @@ struct wildmatch_data {
 };
 
 W_CAP_REG("wildmatch")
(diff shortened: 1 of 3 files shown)


sid0:
Pushed as 532c5f9f89cbd634a09f78a5fcae66e6cff38e25, thanks!

PR #313, opened 2016-07-26
Adding the possibility to exclude file patterns from watchman-make

cotizo:
This commit is adding the `-e PATTERN [PATTERN]` (or `--exclude`) command-line argument to watchman-make. Using it results in files matching the `PATTERN` being ignored by watchman, and thus not triggering a build/make. Excludes have higher priority than matches (i.e. in case file `FOO` matches both `-p PAT` and `-e EXC`, it will in fact be ignored).

### DIFF
--- a/python/bin/watchman-make
+++ b/python/bin/watchman-make
@@ -7,13 +7,19 @@ import sys
 import subprocess
 
 
-def patterns_to_terms(pats):
+def patterns_to_terms(pats, excl):
     # convert a list of globs into the equivalent watchman expression term
-    if pats is None or len(pats) == 0:
-        return ['true']
-    terms = ['anyof']
-    for p in pats:
-        terms.append(['match', p, 'wholename', {'includedotfiles': True}])
+    terms = ['allof', ['type', 'f']]
+    if pats:
+      p_terms = ['anyof']
+      for p in pats:
+          p_terms.append(['match', p, 'wholename', {'includedotfiles': True}])
+      terms.append(p_terms)
+    if excl:
+      e_terms = ['anyof']
+      for e in excl:
+          e_terms.append(['match', e, 'wholename', {'includedotfiles': True}])
+      terms.append(['not', e_terms])
     return terms
 
 
@@ -27,20 +33,21 @@ class Target(object):
     we should execute the command to build this target.
     """
 
-    def __init__(self, name, make, targets, patterns):
+    def __init__(self, name, make, targets, patterns, excluded_patterns):
         self.name = name
         self.make = make
         self.targets = targets
         self.patterns = patterns
+        self.excluded_patterns = excluded_patterns
         self.triggered = False
 
     def __repr__(self):
-        return '{make=%r targets=%r pat=%r}' % (
-            self.make, self.targets, self.patterns)
+        return '{make=%r targets=%r pat=%r excl=%r}' % (
+            self.make, self.targets, self.patterns, self.excluded_patterns)
 
     def start(self, client, root):
         query = {
-            'expression': patterns_to_terms(self.patterns),
+            'expression': patterns_to_terms(self.patterns, self.excluded_patterns),
             'fields': ['name']
         }
         watch = client.query('watch-project', root)
@@ -53,9 +60,23 @@ class Target(object):
         # get the initial clock value so that we only get updates
         query['since'] = client.query('clock', root_dir)['clock']
 
-        print('# Changes to files matching %s will execute `%s %s`' % (
-            ' '.join(self.patterns), self.make, ' '.join(self.targets)),
-            file=sys.stderr)
+        configuration_message = ''
+        if not(not self.patterns):
+            configuration_message += 'matching %s' % (' '.join(self.patterns))
+        if not(not self.excluded_patterns):
+            if len(configuration_message) > 0:
+                configuration_message += ', but '
+            configuration_message += 'not matching %s' % (' '.join(self.excluded_patterns))
+        if len(configuration_message) > 0:
+            print('# Changes to files %s will execute `%s %s`' % (
+                configuration_message,
+                self.make,
+                ' '.join(self.targets)), file=sys.stderr)
+        else:
+            print('# Changes to any file will execute `%s %s`' % (
+                self.make,
+                ' '.join(self.targets)), file=sys.stderr)
+
         sub = client.query('subscribe', root_dir, self.name, query)
 
     def consumeEvents(self, client):
@@ -88,17 +109,13 @@ class DefineTarget(argparse.Action):
         if isinstance(values, basestring):
             values = [values]
 
-        if namespace.pattern is None or len(namespace.pattern) == 0:
-            print('no patterns were specified for target %s' %
-                  values, file=sys.stderr)
-            sys.exit(1)
-
         target = Target('target_%d' % len(targets),
-                        namespace.make, values, namespace.pattern)
+                        namespace.make, values, namespace.pattern, namespace.exclude)
         targets.append(target)
 
         # Clear out patterns between targets
         namespace.pattern = None
+        namespace.exclude = None
 
 
 parser = argparse.ArgumentParser(
@@ -145,6 +162,14 @@ parser.add_argument('--root', type=str, default='.', help="""
 Define the root of the project.  The default is to use the PWD.
 All patterns are considered to be relative to this root, and the build
 tool is executed with this location set as its PWD.
+""")
+parser.add_argument('-x', '--exclude', type=str, nargs='+',
+                    help="""
+Define a set of filename patterns which, if matching, will mark files as
+ignored from the next --target definition.
+
+The pattern syntax is wildmatch style; globbing with recursive matching via
+'**'. Exclude patterns have higher priority than match patterns.
 """)
 
 args = parser.parse_args()
--- a/python/bin/watchman-make
+++ b/python/bin/watchman-make
@@ -171,6 +171,9 @@ ignored from the next --target definition.
 The pattern syntax is wildmatch style; globbing with recursive matching via
 '**'. Exclude patterns have higher priority than match patterns.
 """)
+parser.add_argument('-U', '--sockname', type=str, default=None, help="""
+Specify the path to the UNIX socket used to communicate with Watchman.
+""")
 
 args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @wez to be a potential reviewer.

python/bin/watchman-make
```diff
@@ -7,13 +7,19 @@ import sys
...
+def patterns_to_terms(pats, excl):
     # convert a list of globs into the equivalent watchman expression term
     if pats is None or len(pats) == 0:
         return ['true']
-    terms = ['anyof']
+    p_terms = ['anyof']
     for p in pats:
-        terms.append(['match', p, 'wholename', {'includedotfiles': True}])
+        p_terms.append(['match', p, 'wholename', {'includedotfiles': True}])
+    if excl is None or len(excl) == 0:
```
wez: I believe that you can write this more succinctly as:

```
if not excl:
    return p_terms
```

facebook-github-bot-7:
@cotizo updated the pull request.

python/bin/watchman-make
```diff
@@ -146,6 +159,14 @@ Define the root of the project.  The default is to use the PWD.
 All patterns are considered to be relative to this root, and the build
 tool is executed with this location set as its PWD.
 """)
+parser.add_argument('-e', '--exclude', type=str, nargs='+',
```
wez: can we change `-e` to `-x` or `-X`?  This is slightly more consistent with other unix tools and our own legacy simple pattern syntax (https://facebook.github.io/watchman/docs/simple-query.html).  I'm aware that the `-p` option above isn't consistent with the simple pattern syntax either, but it's not worth changing.

`-e` makes me think of executing rather than excluding.

wez:
Can you include in this PR and the final commit message some examples of combinations of `-p` and `-e` (or `-x`) and the out from the command?  That at least gives the next person that modifies this code a manual test plan in the absence of test coverage!

python/bin/watchman-make
```diff
@@ -7,13 +7,19 @@ import sys
 import subprocess
 
 
-def patterns_to_terms(pats):
+def patterns_to_terms(pats, excl):
     # convert a list of globs into the equivalent watchman expression term
-    if pats is None or len(pats) == 0:
+    if not pats:
         return ['true']
```
wez: probably also want to apply exclusions in this case?

facebook-github-bot-7:
@cotizo updated the pull request.

cotizo:
I wouldn't mind adding test coverage, but I will need to check out first how the project organizes the tests. :)

wez:
the tricky part about testing here is that `watchman-make` and `watchman-wait` are self-contained python scripts (aside from their `pywatchman` dep).  It would be pretty easy to factor out the code to unit test it (we use regular python tests), but it would make more work in our FB-specific deployment for these utilities (and that is slightly different for Linux and Mac).

If you want to add coverage, it may be easier (and honestly a bit more valuable) to make an integration test that launches these tools and then matches some output from it.  I'd suggest using something like `tests/integration/test_trigger.py` as a model and then using the `subprocess` to launch `watchman-make --make /bin/echo`.  The stdout from the process will then be the list of target names that get triggered.

Concurrently with that launched process, you'd then want to touch/change/delete files to trigger things, and you can then make some test assertions on the output from the child process.

wez:
You can use `./runtests.py tests/integration/test_SOMETHING.py` to run just a single test.  There are some options to the `runtests.py` script like `--keep-if-fail` that are useful to debug the isolated watchman process logs used during the test run.  That's handy if you redirect the output from `watchman-make` to a temporary file during the test rather than just using a pipe, because it allows you to post-mortem inspect the full output more easily.

facebook-github-bot-7:
@cotizo updated the pull request.

facebook-github-bot-7:
@cotizo updated the pull request.

cotizo:
Now, you can either specify any combination of patterns to include (possibly none) or to exclude (possibly none). 

* No filters:

```
./watchman-make -t foo
# Relative to /Users/cotizo/watchman/python/bin
# Changes to any file will execute `make foo`
# waiting for change
```

* Filtering on included patterns:

```
./watchman-make -p **/.bar -t foo
# Relative to /Users/cotizo/watchman/python/bin
# Changes to files matching **/.bar will execute `make foo`
# waiting for changes
```

* Filtering on excluded patterns:

```
./watchman-make -x **/.baz -t foo
# Relative to /Users/cotizo/watchman/python/bin
# Changes to files not matching **/.baz will execute `make foo`
# waiting for changes
```

* Filtering on both included and excluded patterns:

```
./watchman-make -p **/.bar -x **/.baz -t foo
# Relative to /Users/cotizo/watchman/python/bin
# Changes to files matching **/.bar, but not matching **/.baz will execute `make foo`
# waiting for changes
```

Not specifying any `-p PATTERN` is not recommended, as any file (e.g. hidden files, short-lived files used by VCS or other processes) will trigger a make.

`watchman-make` is now limited to files only because, if the user uses only `-x PATTERN` without any `-p PATTERN`, then touching any file `//FOO/BAR` will also mark the folder `//FOO/` as dirty, which will in turn trigger a make.

facebook-github-bot-7:
@cotizo updated the pull request.

facebook-github-bot-7:
@cotizo updated the pull request.

wez:
Regarding the files restriction thing, I think you're working around https://github.com/facebook/watchman/issues/305
Can you try applying the fix from there and see if you need that?  I'd rather not add the files constraint to the expression if we can avoid it.

facebook-github-bot-6:
@cotizo updated the pull request.

facebook-github-bot-6:
@cotizo updated the pull request.

facebook-github-bot-6:
@cotizo updated the pull request.

facebook-github-bot-6:
@cotizo updated the pull request.

facebook-github-bot-6:
@cotizo updated the pull request.

cotizo:
Given that most of the CI errors are not related to this PR, and that #305 has already been merged, I would assume we are ready to go. :-)

python/bin/watchman-make
```diff
@@ -146,6 +163,17 @@ Define the root of the project.  The default is to use the PWD.
...
 """)
+parser.add_argument('-x', '--exclude', type=str, nargs='+',
+                    help="""
+Define a set of filename patterns which, if matching, will mark files as
+ignored from the next --target definition.
+
+The pattern syntax is wildmatch style; globbing with recursive matching via
+'**'. Exclude patterns have higher priority than match patterns.
+""")
+parser.add_argument('-U', '--sockname', type=str, default=None, help="""
```
wez: For testing, we normally just set `WATCHMAN_SOCK` in the environment; pywatchman will pick that up

tests/integration/test_watchman_make.py
```diff
@@ -0,0 +1,114 @@
...
+import subprocess
+import signal
+import time
+import uuid
+
+class TestWatchmanMake(WatchmanTestCase.WatchmanTestCase):
+
+    def setUp(self):
+        if sys.platform == 'win32':
+            self.skipTest('Continuous integration testing only in Linux / MacOS')
```
wez: is this just because of `/bin/echo`?

tests/integration/test_watchman_make.py
```diff
@@ -0,0 +1,114 @@
...
+        if sys.version_info >= (3, 0):
+            self.skipTest('watchman-make is only compatible with Python 2')
+        if self.transport != 'local' or self.encoding != 'bser':
+            self.skipTest('Testing only with local transport and bser encoding')
+        self.marker = 'target-' + str(uuid.uuid4())
+
+    def _run_test(self, args, behavior):
+        root = self.mkdtemp()
+        self.touchRelative(root, '.watchmanconfig')
+        fake_env = dict(os.environ)
```
wez: you can set `fake_env['WATCHMAN_SOCK'] = self.getClient().sockpath` here and avoid adding the `-U` option to `watchman-make`

tests/integration/test_watchman_make.py
```diff
@@ -0,0 +1,114 @@
...
+        behavior(root)
+        process.send_signal(signal.SIGINT)
+        return (process.stdout.read(), process.stderr.read())
+
+    def test_noParams(self):
+        def behavior(root):
+            self.touchRelative(root, "foo")
+            time.sleep(1)
+        (stdout, stderr) = self._run_test([], behavior)
+        print(stdout)
```
wez: stray prints?

tests/integration/test_watchman_make.py
```diff
@@ -0,0 +1,114 @@
...
+            '--make', '/bin/echo']
+        cmdline.extend(args)
+        cmdline.extend(['-t', self.marker])
+        process = subprocess.Popen(
+            ' '.join(cmdline),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True,
+            env=fake_env)
+        time.sleep(1)
```
wez: fixed duration sleeps tend to result in flaky tests and make tests take longer to run.

We have `WatchmanTestCase.waitFor` to help poll a condition in a loop with a timeout; this helps to notice when the condition is true sooner and also gives us some slack if the CI is overloaded.

Could you take another pass over this test and see if you could make use of `waitFor` or `assertWaitFor`?

wez:
Thanks for adding tests!  Would you mind tweaking them to avoid the sleeps?

tests/integration/test_watchman_make.py
cotizo: Oh derp, I forgot about `/bin/echo`... but no, it is because I can't sigint on windows, and I have to send a subprocess.CTRL_C_EVENT, which is available only on windows. I had a commit with that, but it still failed, so I decided to run the tests only under Unixes, where subprocess acts nicely. In the end, watchman-make will still work on Windows, it's only the testing (spawning processes) which is much easier on *nix. :-)

tests/integration/test_watchman_make.py
cotizo: Will give it a shot, but the problem is that I would have to consume the stdout/stderr as streams (as opposed to strings), which is a bit harder. :) 

About the `time.sleep`s from the `behavior` methods passed from each test: those should still be sleep commands, but with lower timeouts. (say, 2x the settle period) I'll just use `waitFor` with an always-false lambda though.

wez:
This has been lingering for a while.  if/when you have the bandwidth, please adjust per the review comments and re-submit.  Thanks!

atrefz:
Is this tracked internally somehow by any chance?

This seems like rather useful & important functionality - it would be awesome if the last few revisions to the tests could be made to land this - I would be happy to try myself, but i doubt my python is solid enough to fix the issues that have been raised here.

wez:
No one is working on this and we'd welcome your efforts to finish it up!

PR #314, opened 2016-07-27
Fixing watchman.jar to include dependencies

cotizo:
Building the jar without bundling dependencies is confusing, since users would have to manually include them to prevent `ClassNotFoundException`. This PR changes the buck target type, so all the classes of our dependencies are included in the final jar. 

Bonus: watchman.jar is produced in a much friendlier location (`buck-out/gen/watchman.jar`).

### DIFF
--- a/java/BUCK
+++ b/java/BUCK
@@ -1,5 +1,12 @@
-java_library(
+java_binary(
   name = 'watchman',
+  deps = [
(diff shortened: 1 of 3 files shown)


facebook-github-bot-9:
@cotizo updated the pull request.

wez:
Thanks!

PR #316, opened 2016-07-28
Watchman Python3/Asyncio Client

dhruvsinghal:


facebook-github-bot-1:
By analyzing the blame information on this pull request, we identified @wez, @sid0 and @otrempe to be potential reviewers.

facebook-github-bot-3:

facebook-github-bot-3:

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env python3
+
+from __future__ import print_function
```
ambv: nit: That's not required for Python 3-only code.

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,222 @@
...
+import asyncio
+import pywatchman
+import os
+import sys
+
+def fieldlist(s):
+    # helper for splitting a list of fields by comma
+    return s.split(',')
+
+parser = argparse.ArgumentParser(
```
ambv: Typically you'd not mix function definitions with module-level executed code. Rather, you'd define your functions and create a main() function, that you'd call at the end using a so-called `ifmain` block:
```
if __name__ == '__main__':
    main()
```

That way we can unit test the functions in the module by importing it from somewhere else without running the script code accidentally.

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,222 @@
...
+    relpath = None  # Offset to dir of interest
+    name = None     # Our name for this subscription
+    path = None
+
+    def __init__(self, name, path):
+        self.name = name
+        self.path = path
+
+    def __repr__(self):
+        return 'Subscription(root=%s, rel=%s, name=%s)' % (
```
ambv: The (soft) contract of `__repr__` is to be able to recreate the object by executing the representation. So it'd be nice if `__init__` accepted `rel`, better yet named `relpath`, just like the field.

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,222 @@
...
+            subscriptions.items()]
+    loop.run_until_complete(asyncio.gather(*tasks, loop=loop))
+
+except pywatchman.CommandError as ex:
+    print('watchman:', ex.msg, file=sys.stderr)
+    sys.exit(1)
+
+except KeyboardInterrupt:
+    # suppress ugly stack trace when they Ctrl-C
+    sys.exit(3)
```
ambv: You'll still get warnings from asyncio about not closing the event loop and pending tasks. You should add a finally: block that cancels pending tasks, joins them and closes the loop.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,286 @@
...
+    '''Local Unix domain socket transport supporting asyncio.'''
+
+    def __init__(self, sockname, loop):
+        self.sockname = sockname
+        self.reader = None
+        self.writer = None
+        init_task = asyncio.ensure_future(self.__init_socket(), loop=loop)
+        loop.run_until_complete(init_task)
+
+    async def __init_socket(self):
```
ambv: nit: a single leading underscore is enough.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,286 @@
...
+
+class AsyncUnixSocketTransport(AsyncTransport):
+    '''Local Unix domain socket transport supporting asyncio.'''
+
+    def __init__(self, sockname, loop):
+        self.sockname = sockname
+        self.reader = None
+        self.writer = None
+        init_task = asyncio.ensure_future(self.__init_socket(), loop=loop)
+        loop.run_until_complete(init_task)
```
ambv: That means you can't instantiate new AUSTs from within a running event loop. For this case, provide a class method constructor (like the `from_path()` in the other file) that is a coroutine and just await on the init task.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,286 @@
...
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import asyncio
+import os
+import subprocess
```
ambv: There is an asyncio subprocess equivalent.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,286 @@
...
+import threading
```
ambv: Doesn't seem you're using this.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,286 @@
...
+            await self.logs.put(response['log'])
+
+        elif 'subscription' in response:
+            sub = response['subscription']
+            root = os.path.normcase(response['root'])
+            self.__ensure_subscription_queue_exists(sub, root)
+            await self.sub_by_root[root][sub].put(response)
+
+        elif self.__is_unilateral(response):
+            raise Exception('Unknown unilateral response.')
```
ambv: Prefer creating a named exception class for your exceptions. That way your clients can catch them specifically and don't need to resort to Pokemon exception handling.

dhruvsinghal: Fixed. Changed to WatchmanError.

python/pywatchman/aioclient.py
dhruvsinghal: Fixed.

python/pywatchman/aioclient.py
dhruvsinghal: I actually created a class method constructor that does the initialization instead. The constructor is now more decoupled from specifics of the Transports and Codecs.

python/pywatchman/aioclient.py
dhruvsinghal: https://github.com/python/asyncio/issues/281.

As discussed offline, I still create a regular subprocess, but use submit it to a threadpool to make it asynchronous.

python/bin/watchman-wait-aio
dhruvsinghal: Removed.

facebook-github-bot-7:

facebook-github-bot-3:

python/bin/watchman-wait-aio
dhruvsinghal: So, following your comment, I have heavily restructured the `watchman-wait` code to make it more modular.

python/bin/watchman-wait-aio
dhruvsinghal: I changed the `__repr__` to `__str__`, but also made `__init__` take relpath. There also are other changes due to restructuring, so please take a look.

python/bin/watchman-wait-aio
dhruvsinghal: Fixed this.

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,241 @@
...
+
+        query = {
+            'expression': expr,
+            'fields': fields,
+        }
+
+        watch = await client.query('watch-project', dir_to_watch)
+
+        if 'warning' in watch:
+            print('WARNING: ', watch['warning'], file=sys.stderr)
```
ambv: nit: Since this is going to be a library used within external applications, using Python's `logging` module would be better.

python/bin/watchman-wait-aio
```diff
@@ -0,0 +1,241 @@
...
+
+    async def get(self, client):
+        data = await client.get_subscription(self.name, self.root)
+        return data['files']
+
+    @classmethod
+    def from_path(cls, path, name=None):
+        '''Create a subscription for the given path, using that path as the
+        subscription's name if not specified.'''
+        name = name if name is not None else path
```
ambv: nit: not important for now but in Python you can do `name = name or path` which matches your case well enough and is easier to read :)

tests/async/AsyncWatchmanTestCase.py
```diff
@@ -0,0 +1,97 @@
...
+    def setUp(self):
+        self.loop = asyncio.get_event_loop()
+        self.client = self.loop.run_until_complete(WatchmanClient.from_socket())
+
+    def tearDown(self):
+        self.client.close()
+
+    def run(self, result):
+        if result is None:
+            raise Exception('MUST be a runtests.py:Result instance')
```
ambv: nit: You could just `assert result and isinstance(result, Result)`

tests/async/AsyncWatchmanTestCase.py
```diff
@@ -0,0 +1,97 @@
...
+        super(AsyncWatchmanTestCase, self).run(result)
+        return result
+
+    def mkdtemp(self):
```
ambv: What's the point of this method? You're importing `tempfile` in the test case file anyway.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+    def _check_error(self, res):
+        if 'error' in res:
+            raise CommandError(res['error'])
+
+    def _check_recieve_loop(self):
+        if self.receive_task is None:
+            raise ReceiveLoopError('Recieve loop was not started.')
+
+    def _is_unilateral(self, res):
+        return 'unilateral' in res and res['unilateral'] or \
```
ambv: nit: `res.get('unilateral')` is the same as `'unilateral' in res and res['unilateral']`

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+            await self.logs.put(response['log'])
+
+        elif 'subscription' in response:
+            sub = response['subscription']
+            root = os.path.normcase(response['root'])
+            self._ensure_subscription_queue_exists(sub, root)
+            await self.sub_by_root[root][sub].put(response)
+
+        elif self._is_unilateral(response):
+            raise WatchmanError('Unknown unilateral response.')
```
ambv: You could add the keys in the response to the error for the user to be able to identify what that response contained. Or the entire response, actually, it might be helpful for debugging.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+
+        else:
+            raise WatchmanError('Not a unilateral response.')
```
ambv: Ditto here.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+    async def read(self, size):
+        res = await self.reader.read(size)
+        return res
+
+    def close(self):
+        if self.writer:
+            self.writer.close()
+
+
+class AsyncCodec(object):
```
ambv: nit: leave this as-is but in the future keep in mind that in Python "codec" is a protocol that contains the methods `encode` and `decode`. `send`/`receive` implies connections, most often sockets. Funnily, you call the instantiated AsyncBserCodec in line 284 a `connection` yourself :)

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+        except ValueError as e:
+            raise WatchmanError('watchman response decode error: %s' % e)
+
+    async def send(self, *args):
+        cmd = bser.dumps(*args) # Defaults to BSER v1
+        await self.transport.write(cmd)
+
+    def _loads(self, response):
+        ''' Parse the BSER packet '''
+        return bser.loads(response) # Defaults to BSER v1
```
ambv: How big can a request (dumps in line 179) and a response be? The loads/dumps API is inherently hostile towards asyncio. That's a general problem, also true for JSON, pickle, etc. etc. As long as the objects are small, we can ignore this. If the objects can be big (e.g. take up more than a second to serialize/deserialize), we should TODO an improvement here (for example a thread pool executor if it frees the GIL at any point of ser/deser).

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+            raise WatchmanError('empty watchman response')
+
+        _1, _2, elen = bser.pdu_info(buf[0])
+
+        rlen = len(buf[0])
+        while elen > rlen:
+            buf.append(await self.transport.read(elen - rlen))
+            rlen += len(buf[-1])
+
+        response = b''.join(buf)
```
ambv: For known expected length, prefer allocating bytearrays instead of `b''.join`. The former is faster, which will add up in a `receive()` method over time.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+            if self._is_unilateral(response):
+                await self._process_unilateral_response(response)
+            else:
+                await self.bilateral_response_queue.put(response)
+
+    def _check_error(self, res):
+        if 'error' in res:
+            raise CommandError(res['error'])
+
+    def _check_recieve_loop(self):
```
ambv: Typo: the word is `receive` (`e` and `i` are flipped).

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+    def close(self):
+        '''Close the underlying connection.'''
+        if self.receive_task:
+            self.receive_task.cancel()
+        if self.connection:
+            self.connection.close()
+
+    def enable_receiving(self, loop=None):
+        '''Schedules the receive loop to run on the given loop.'''
+        self.receive_task = asyncio.ensure_future(self._receive_loop(),
```
ambv: You should schedule a callback for an exception happening in this task. It's an endless while loop that can't be broken if you cancel it (that's expected) or if you get any response that raises WatchmanError (that's unlikely but we should support recovery here). asyncio will be unhappy with you if there was an exception stored in the task and it was never read by your program.

Also, this doesn't seem like a tested path in the client's tests.

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,340 @@
...
+    def _check_recieve_loop(self):
+        if self.receive_task is None:
+            raise ReceiveLoopError('Recieve loop was not started.')
+
+    def _is_unilateral(self, res):
+        return 'unilateral' in res and res['unilateral'] or \
+               'subscription' in res or \
+               'log' in res
+
+    def _ensure_subscription_queue_exists(self, name, root):
```
ambv: This can be achieved in Python by:
`self.sub_root.setdefault(root, {}).setdefault(name, asyncio.Queue())`

runtests.py
```diff
@@ -16,6 +16,7 @@
 # Ensure that we can find pywatchman
 sys.path.insert(0, os.path.join(os.getcwd(), 'python'))
 sys.path.insert(0, os.path.join(os.getcwd(), 'tests', 'integration'))
+sys.path.insert(0, os.path.join(os.getcwd(), 'tests', 'async'))
```
ambv: Maybe don't add it to path if Python isn't 3.5+?

python/bin/watchman-wait-aio
dhruvsinghal: If I recall correctly, `watchman-wait` (and, hence, `watchman-wait-aio`) is a stand-alone application. @sid0, @wez is there some library-like use for this program?

python/bin/watchman-wait-aio
dhruvsinghal: Done.

tests/async/AsyncWatchmanTestCase.py
dhruvsinghal: Done.

tests/async/AsyncWatchmanTestCase.py
dhruvsinghal: Removed it. It was left over from the old `WatchmanTestCase`.

python/pywatchman/aioclient.py
dhruvsinghal: Fixed.

python/pywatchman/aioclient.py
dhruvsinghal: Fixed.

python/pywatchman/aioclient.py
dhruvsinghal: Fixed.

python/pywatchman/aioclient.py
dhruvsinghal: @sid0, @wez: What do you think?

python/pywatchman/aioclient.py
dhruvsinghal: 👍

python/pywatchman/aioclient.py
dhruvsinghal: Fixed.

runtests.py
dhruvsinghal: Done.

python/bin/watchman-wait-aio
sid0: `watchman-wait` is a program, so `print` is fine here.

python/pywatchman/aioclient.py
sid0: I think we could TODO an improvement here, but at the moment it won't really be possible to release the GIL in the middle of the `loads` -- the BSER deserializer tends to create Python objects on the fly.

python/pywatchman/aioclient.py
dhruvsinghal: Done.

python/pywatchman/aioclient.py
dhruvsinghal: Done.

python/pywatchman/aioclient.py
dhruvsinghal: Done.

facebook-github-bot-0:
@dhruvsinghal updated the pull request - [view changes](https://github.com/facebook/watchman/pull/316/files/5f0ab0f89f7b5d538b489cd5565ae6d1fc2a10c3..2db37d17fa967e6bfa39dd16bb53ac6a699b58d2)

python/pywatchman/aioclient.py
ambv: Are those requests and responses typically small though?

python/pywatchman/aioclient.py
```diff
@@ -0,0 +1,347 @@
...
+    def enable_receiving(self, loop=None):
+        '''Schedules the receive loop to run on the given loop.'''
+        self.receive_task = asyncio.ensure_future(self._receive_loop(),
+                                                  loop=loop)
+        def do_if_done(fut):
+            try:
+                fut.result()
+            except CancelledError:
+                pass
+            except ex:
```
ambv: Did you meant `except Exception as ex:`?

python/pywatchman/aioclient.py
sid0: Depends entirely on what's asked for. If you request data for 100k files then it'll be biggish (~10-30MB).

ambv: Alright, that seems like a decent future improvement to make then. Out of curiosity, does Watchman prefer a few large subscriptions or many small subscriptions?

tests/async/test_subscribe_async.py
```diff
@@ -0,0 +1,75 @@
...
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+import pywatchman
+
+
+class TestSubscribe(AsyncWatchmanTestCase.AsyncWatchmanTestCase):
+
+    def test_subscribe(self):
```
ambv: I still think a test for an exception case (mocked or a real empty response, I don't care) would be nice. You'd catch the `except ex:` NameError that way.

wez:
What's left to do here to get this merged?

ambv:
@wez, there's two outstanding questions I had here but I think it's worth just merging what we have and fixing forward.

indygreg:
Per last comment, this appears "worth merging."

@sid0 is there anything I can do to help get an asyncio client landed?

PR #317, opened 2016-07-31
Shave off some memory usage around watchman_file

wez:
Two commits here:

1. We make better usage of the allocator bucketing/binning by embedding the filename string directly into the watchman_file struct

2. We never use more than second granularity for timestamp based (rather than tick based) since comparisons, so we can save 16 bytes per file entry if we switch from using a timeval to a time_t.

### DIFF
--- a/clockspec.c
+++ b/clockspec.c
@@ -47,8 +47,7 @@ struct w_clockspec *w_clockspec_parse(json_t *value) {
 
   if (json_is_integer(value)) {
     spec->tag = w_cs_timestamp;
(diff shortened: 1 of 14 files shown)


facebook-github-bot-2:

facebook-github-bot-7:

sid0:
LGTM

wez:
Pushed as f77642393e318388225d82516935c4e57fd6b384 and 8c11be824783801dbed27d2e6f136939f0fcf14e

PR #318, opened 2016-07-31
Reduce memory utilization around dir names

wez:
We've traditionally maintained a map of full dir name -> dir structure.
As the volume and depth of a watched tree increases, this results in
a lot of duplicate path prefixes maintained in memory, bloating the
watchman process.

This diff trades the O(1) lookup time complexity for O(depth-of-path-in-root)
for a reduced memory footprint; we're now storing just a single path
component in each directory and file.  This means that we have to
introduce some helper functions to compute the effective directory
path on demand.

This translates into about 40% memory savings (but see note below)
and should be more on OSX.  I've only done some light testing with
this because my local repos are more sparse than is representative; I
will look at this when I'm plugged into the ethernet.

Note that inotify and kqueue watches maintain additional maps keyed by
dirname; this diff does not change that, and there is potentially
more to be saved by looking at that in more detail

(Note: this PR is stacked on top of https://github.com/facebook/watchman/pull/317 so you'll see a couple of commits from that in here)

### DIFF
--- a/cmds/debug.c
+++ b/cmds/debug.c
@@ -107,7 +107,6 @@ static void cmd_debug_poison(struct watchman_client *client, json_t *args)
 {
   w_root_t *root;
   struct timeval now;
-  struct watchman_dir *dir;
   json_t *resp;
 
   root = resolve_root_or_err(client, args, 1, false);
@@ -115,11 +114,9 @@ static void cmd_debug_poison(struct watchman_client *client, json_t *args)
     return;
   }
 
-  dir = w_root_resolve_dir(root, root->root_path, false);
-
   gettimeofday(&now, NULL);
 
-  set_poison_state(root, dir->path, now, "debug-poison", ENOMEM, NULL);
+  set_poison_state(root, root->root_path, now, "debug-poison", ENOMEM, NULL);
 
   resp = make_response();
   set_unicode_prop(resp, "poison", poisoned_reason);
--- a/string.c
+++ b/string.c
@@ -4,6 +4,22 @@
 #include "watchman.h"
 #include <stdarg.h>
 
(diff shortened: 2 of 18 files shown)


facebook-github-bot-6:

facebook-github-bot-7:

facebook-github-bot-7:

facebook-github-bot-7:

cmds/debug.c
```diff
@@ -107,19 +107,16 @@ static void cmd_debug_poison(struct watchman_client *client, json_t *args)
...
   struct timeval now;
-  struct watchman_dir *dir;
   json_t *resp;
 
   root = resolve_root_or_err(client, args, 1, false);
   if (!root) {
     return;
   }
 
-  dir = w_root_resolve_dir(root, root->root_path, false);
```
sid0: LOL

string.c
```diff
@@ -4,6 +4,22 @@
 #include "watchman.h"
 #include <stdarg.h>
 
+/** An optimization to avoid heap allocations during a lookup, this function
+ * creates a string object on the stack.  This object does not own the memory
+ * that it references, so it is the responsibility of the caller
+ * to ensure that that memory is live for the duration of use of this string.
+ * It is therefore invalid to add a reference or take a slice of this stack
+ * string as the lifetime guarantees are not upheld. */
```
sid0: Rust would probably help a lot here ;)

sid0:
I had a look through the diffs and they all make sense. Nothing obviously wrong in them either.

string.c
wez: yeah, I was wishing for either that or C++ here while writing this :-)

wez:
Pushed as b1aea7a

PR #323, opened 2016-08-03
fix windows timeout

otrempe:
Failing to acquire a critical section and then calling a WaitForSingleObject
does not return causing lots of timeouts under Windows.


facebook-github-bot-3:
By analyzing the blame information on this pull request, we identified @wez and @dhruvsinghal to be potential reviewers.

wez:
Oh, great find!  Is the semaphore even present in the cases that it doesn't work?

We have this initializer:

```
#define PTHREAD_MUTEX_INITIALIZER {(void*)-1,-1,0,0,0,0}
```

I don't know if this is 100% equivalent to `InitializeCriticalSection`... if that is the cause of this it may be slightly better to do something like this:

```
struct pthread_mutex {
   pthread_spinlock_t initializer_spin_lock;
   BOOL initialized;
   CRITICAL_SECTION cs;
}
#define PTHREAD_MUTEX_INITIALIZER {PTHREAD_SPINLOCK_INITIALIZER, 0, {0}}

static void ensure_mutex_init(pthread_mutex *m) {
   if (m->initialized) return;
   pthread_spin_lock(&m->initializer_spin_lock);
   if (!m->initialized) {
      InitializeCriticalSection(&m->cs);
      m->initialized = 1;
   }
   pthread_spin_unlock(&m->initializer_spin_lock);
}
```

and then make sure that we're calling `ensure_mutex_init` in each of the `pthread_mutex_XXX` functions.

vmnet4:
Hi wez. I'm working with otrempe on this.
To answer your question, yes the semaphore was present on the failing cases, I mean it was different of 0. I found some article on the web telling that `LockSemaphore` would be in fact an auto-reset event (instead of a semaphore like the variable name suppose). So from what I understand, in theory, the actual code should work. In practice though, I experiment a 60secs timeout on server side as soon as the first `pthread_mutex_trylock` of the function fails to acquire the critical section...
...while replacing the `WaitForSingleObject` by a tiny sleep does the job on the first encountered `pthread_mutex_trylock` inside the while loop:

```
while (1)
  {
    /* Have we waited long enough? */
    if (ct > t) return ETIMEDOUT;

    /* Wait on semaphore within critical section */
    // WaitForSingleObject(((struct _pthread_crit_t *)m)->sem, (DWORD)(t - ct));
    SleepEx( 20, TRUE );

    /* Try to grab lock */
    if (!pthread_mutex_trylock(m)) return 0;

    /* Get current time */
    ct = _pthread_time_in_ms();
  }
```

Concerning what you suggest I'm not sure I get it. Should I understand that you suspect there is something wrong with `InitializeCriticalSection`? `InitializeCriticalSection` is done when calling `w_root_new` and from what I remember of the log files, there was several successful "query command" prior to a failing one using the same watched folder. I will verify tomorrow but if the same root is used across those query requests, calling `ensure_mutex_init` on every `pthread_mutex_trylock` call won't help because `m->initialized` will be true on the actual failing cases, right?

wez:
Thanks for checking and explaining!   I was concerned that we weren't properly initializing the structure in all cases, but it sounds like we're OK there, so you can ignore my suggestion.

It sounds dangerous to try to wait on the internal "semaphore" if it is an auto-reset event, and that means that an unconditional sleep is probably the best we can do.

Can we use a shorter sleep than 20ms?  Does it have any adverse effects?

facebook-github-bot-3:
@otrempe updated the pull request.

vmnet4:
I understood you told about the `pthread_mutex_t` in general. I was only investigating issue around the `root->lock` which is the cause of the timeouts on my side. I was confused because that one was already handled with InitializeCriticalSection.
I think your idea is a good one, I think always using the windows API when playing with CRITICAL_SECTION sounds safer. I've push commits in that way but a quick look at the CI makes me think I'll rollback!
Otherwise for your timeout related question, 20ms is totally arbitrary. My feeling is that a really small value will tend to act like an active loop. Though I've tried on my side with 1ms and it seemed ok.
My latest commit was setting the timeout to 5ms...

By the way, I had to remove the calls to `pthread_spin_lock` and `pthread_spin_unlock` because it stalled on the inner while loop on my side. Can you explain me the goal behind those calls?

wez:
The spin locks were just to ensure that only one thread would initialize the critical section (we need a mutex to guard the mutex initialization!)

Sounds like we also have a problem with the spinlock implementation in this library... I don't think we're using it for anything today anyway, so I wouldn't worry too much about it.

I do have plans to flip us over to C++ so that we can gradually start to move over to using C++11 and higher compiler features that should help us avoid these sorts of things in the future, so I don't think we should work too hard to solve these compatibility issues: let's just get something that works :-)

winbuild/pthread.c
```diff
@@ -69,6 +69,21 @@ static void _pthread_once_cleanup(pthread_once_t *o)
   *o = 0;
 }
 
+/* Ensure the CriticalSection has been initialized */
+static void ensure_mutex_init(pthread_mutex_t *m) {
```
wez: inline this?

winbuild/pthread.c
```diff
@@ -69,6 +69,21 @@ static void _pthread_once_cleanup(pthread_once_t *o)
...
+   if (m->initialized) return;
+
+   if (!m->initialized) {
+      InitializeCriticalSection(&m->cs);
+      m->initialized = 1;
+   }
+}
+
+static LPCRITICAL_SECTION pthread_mutex_cs_get(pthread_mutex_t *m) {
```
wez: inline here too?

winbuild/pthread.c
```diff
@@ -69,6 +69,21 @@ static void _pthread_once_cleanup(pthread_once_t *o)
   *o = 0;
 }
 
+/* Ensure the CriticalSection has been initialized */
+static void ensure_mutex_init(pthread_mutex_t *m) {
+   if (m->initialized) return;
+
+   if (!m->initialized) {
```
wez: this is racy; if two threads are calling in here at the same time, one may initialize and lock the critical section before the other initializes and then successfully locks it; that was the reason behind spinlocking.

An alternative to fixing up the spinlock is using a constructor function to initialize a global mutex in here just for guarding this bit of code.

I can't comment more right now, but take a look at `w_ctor_fn_reg` for registering a constructor

facebook-github-bot-3:
@otrempe updated the pull request.

vmnet4:
I rollbacked, but let the 5ms instead of 20ms. I think that's fine for the 3s timeout problem on windows.

Whatever I try to keep the `ensure_mutex_init` get me stuck into a deadlock while running the tests.
Going in C++ would definitively be nice since it would allow easy use of initializer lists and the RAII idiom.

Also I could use tools we have in here to find potential problem while playing with thread. That could point out some uninitialized variables or race conditions...

For now my priority will be on a couple of failling tests on my side which are successful on the CI and on my co-worker Win8 OS... Sounds like there are still issues specific to windows 7.

vmnet4:
I expected the windows CI to get back to 2 failed instances. Instead, last run showed 20 fails!?
Shuffle in the test cases order maybe?

facebook-github-bot-3:
@otrempe updated the pull request.

facebook-github-bot-3:
@otrempe updated the pull request.

listener.c
```diff
@@ -8,7 +8,7 @@
 
 /* This needs to be recursive safe because we may log to clients
  * while we are dispatching subscriptions to clients */
-pthread_mutex_t w_client_lock;
+pthread_mutex_t w_client_lock = PTHREAD_MUTEX_INITIALIZER;
```
wez: technically this one needs a recursive initializer, but its okay because we guarantee initialization below (I'll add a line note in a moment).
Please remove this line from this PR as it doesn't contribute to solving the root of the problems :-)

winbuild/pthread.c
```diff
@@ -139,32 +155,36 @@ int _pthread_once_raw(pthread_once_t *o, void (*func)(void))
...
-  return TryEnterCriticalSection(m) ? 0 : EBUSY;
+  return TryEnterCriticalSection(pthread_mutex_cs_get(m)) ? 0 : EBUSY;
 }
 
 int pthread_mutex_init(pthread_mutex_t *m, pthread_mutexattr_t *a)
 {
   (void) a;
-  InitializeCriticalSection(m);
-
+  m->initializer_spin_lock = 0;
```
wez: a bit of a nit, but this is probably clearer this way:

```
InitializeCriticalSection(&m->cs);
m->initialized = TRUE;
m->initializer_spin_lock = 0;
```

winbuild/pthread.c
```diff
@@ -1084,6 +1103,11 @@ int pthread_spin_destroy(pthread_spinlock_t *l)
 /* No-fair spinlock due to lack of knowledge of thread number */
 int pthread_spin_lock(pthread_spinlock_t *l)
 {
+  if ( *l != 0 && *l != EBUSY )
+  {
+    w_log( W_LOG_FATAL, "Fatal error: spinlock value different from 0 or EBUSY! Smells like an uninitialized spinlock. Deadlock insight.");
```
wez: nit:  "\n" at the end of the string here :)

winbuild/pthread.h
```diff
@@ -80,12 +80,16 @@
 #define PTHREAD_CANCELED ((void *)(intptr_t)0xDEADBEEF)
 
 #define PTHREAD_ONCE_INIT 0
-#define PTHREAD_MUTEX_INITIALIZER {(void*)-1,-1,0,0,0,0}
+#define PTHREAD_MUTEX_CS_INITIALIZER {(void*)-1,-1,0,0,0,0}
```
wez: nit: probably shouldn't make this visible with the prefix `PTHREAD_`.  I'd suggest either folding it into the PTHREAD_MUTEX_INITIALIZER and not defining a separate macro, or giving it an underscore prefix eg: `_PTHREAD_MUTEX_CS_INITIALIZER`

wez:
This is looking great, thanks!
Would you mind tweaking this for the couple of suggestions I had?
I plan on committing this PR to the repo tomorrow.

facebook-github-bot-7:
@otrempe updated the pull request.

otrempe:
Changes requested in code review have been pushed

PR #325, opened 2016-08-04
fix for windows 7 and earlier not supporting GetOverlappedResultEx

otrempe:
Work continued from pull request #306


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @wez, @sid0 and @dhruvsinghal to be potential reviewers.

vmnet4:
@wez According to your comments on the PR #306, we now use the same strategy for pywatchman than for watchman itself.

The strategy is: get rid of GetOverlappedResultEx and implement the same behavior using only CreateEvent, WaitForObjectEx and GetOverlappedResult.

Maintenance effort is therefore minimized and comptability is maximized.

stream_win.c
```diff
@@ -46,78 +63,55 @@ typedef BOOL (WINAPI *get_overlapped_result_ex_func)(
...
-      if (millis <= 0) {
-        // Out of time
-        SetLastError(WAIT_TIMEOUT);
-        return FALSE;
-      }
-    }
-  }
-}
 
-static BOOL WINAPI probe_get_overlapped_result_ex(
```
wez: I'd prefer to keep using the real GetOverlappedResultEx function if is is available.  For testing purposes, how about changing this probe function to check for something like WATCHMAN_WIN7_COMPAT=1 in the environment?

If that is set then we can just set `get_overlapped_result_ex = win7_get_overlapped_result_ex`, otherwise just run the code as is below.

That gives us the possibility of configuring a build matrix in the CI:

https://www.appveyor.com/docs/build-configuration#build-matrix

We can have one configuration with this set and one without it set, thus testing both code paths.
We could also use the same environment variable to force this mode on in the python code too.

You get bonus points and gratitude if you can include the appropriate changes to our appveyor.yml file to enable this as part of this pull request!

facebook-github-bot-7:
@otrempe updated the pull request.

otrempe:
I don't know what's going on with AppVeyor. Jobs seem to lock on a 'chocolateyInstall.ps1' because of an interactive prompt. Wez, can you check that please? I have no idea what to do with this.

wez:
Pull and rebase; I've pushed a commit that seems to help

facebook-github-bot-7:
@otrempe updated the pull request.

appveyor.yml
```diff
@@ -22,3 +22,12 @@ artifacts:
...
     name: watchman
     type: zip
+
+environment:
+  matrix:
+    - WATCHMAN_WIN7_COMPAT: 1
+    - WATCHMAN_WIN7_COMPAT:
+
+platform:
+  - x86
```
wez: I had only planned on supporting 64-bit native systems, and I think much of our win32 build setup is just using 64-bit stuff anyway.   I'm not opposed to taking in patches to keep things working on 32-bit systems, but we won't be putting in any real effort to keeping that working.  If you don't need 32-bit support, it would be nicer on the appveyor infra to remove this platform from the matrix :-)

wez:
Looking great, but appveyor is reporting success for this most recent version, even though there were some failing (flaky) tests :-(

I wonder if the make.bat changes from the other PR are causing this?  Appveyor will run `make integration` (it's in the `test_script` section of `appveyor.yml`) and we want the exit status of that to bubble up

facebook-github-bot-2:
@otrempe updated the pull request.

facebook-github-bot-2:
@otrempe updated the pull request.

stream_win.c
```diff
@@ -46,79 +63,79 @@ typedef BOOL (WINAPI *get_overlapped_result_ex_func)(
...
+    BOOL alertable ) {
   get_overlapped_result_ex_func func;
 
   func = (get_overlapped_result_ex_func)GetProcAddress(
       GetModuleHandle("kernel32.dll"),
       "GetOverlappedResultEx");
 
-  if (!func) {
-    func = win7_get_overlapped_result_ex;
+  if ( getenv( "WATCHMAN_WIN7_COMPAT") || !func ) {
```
wez: would be great to explicitly check for the env var containing "1" so that we don't chose to do this if someone set it to 0.

Just adding ` && getenv("WATCHMAN_WIN7_COMPAT")[0] == '1'` in that conditional is probably good enough if you're feeling too lazy to add a local variable for it :-)

wez:
Finally got around to really reading the windows API portions of this; the PR looks great, I'd just like to see a successful run with the two different environment settings.  I think we may have conflicted with some of the windows build options, so I pushed a commit that I think will help.

Would you mind rebasing and updating this PR so that we kick the CI system again?

Thanks!

facebook-github-bot-0:
@otrempe updated the pull request - [view changes](https://github.com/facebook/watchman/pull/325/files/635358dc111ae21c2673db887bad4b5988876cef..67c3701e17a8a2ca4536f231302ce8e1cd4ee0c9)

wez:
Pushed a very slightly squashed/edited (I fixed the line length of a comment and merged a couple of commits together) version of this to master at fe6ee74.

Thank you very much for working on this!

PR #327, opened 2016-08-09
Update install.markdown

thisisaaronland:
Update install.markdown to note requirement for `python-dev` on Linux machines.

### DIFF
--- a/website/_docs/install.markdown
+++ b/website/_docs/install.markdown
@@ -10,7 +10,7 @@ permalink: docs/install.html
 
 Watchman is known to compile and pass its test suite on:
 
- * <i class="fa fa-linux"></i> Linux systems with `inotify`
+ * <i class="fa fa-linux"></i> Linux systems with `inotify` and `python-dev`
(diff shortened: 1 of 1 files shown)


facebook-github-bot-4:

wez:
We already document the python-dev dependency in this section:
https://facebook.github.io/watchman/docs/install.html#installing-from-source
the section you've proposed a change to is intended to be a quick summary of the systems that we run on and not somewhere were we want to exhaustively list the dependencies.

I'm open to suggestions on how to make this clearer!

thisisaaronland:
Huh, you're right - the dependencies are right there.

That said, my experience was that the first section reads as the dependency list itself and the section section as "what do when you have all your dependencies installed".

TBH, the copy over the `pre` section simply didn't register which is maybe lazy on my part but I have a feeling I am not the only person to behave this way when they're just trying to get something up and running enough to determine if they want to proceed further.

I guess this PR is moot given all that but I would mention that I did not specify the `--with-python` flag and the configure/make scripts went ahead and tried to install the Python bindings anyway which is how we got here in the first place :-)

PR #329, opened 2016-08-11
Add configuration option to suppress recrawl warnings

wez:
At FB we now use the perf reporting hooks to sample this centrally which
makes these user facing warnings unactionable noise, so we want to turn
them off.


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @sid0 and @dhruvsinghal to be potential reviewers.

PR #331, opened 2016-08-15
perf: Allow deferring computing hval in our string type

wez:
Eager hval computation was helpful when strings were guaranteed to be
used as keys in hash lookups, but we now have a number of code paths
where the eager hashing is wasteful of CPU.

This diff trades the `hval` field of the struct for an inline function
that will compute it on demand.

This effectively removes the hash_bytes function from the profile of
some heavy watchman queries.


facebook-github-bot-1:
By analyzing the blame information on this pull request, we identified @dhruvsinghal to be a potential reviewer.

facebook-github-bot-9:

wez:
Pushed as 38a3ada

PR #332, opened 2016-08-15
perf: capture the slow query in the query perf sample

wez:
I thought we were already doing this, but apparently not.
This captures a reference to the incoming query so that we can send it
out along with the perf sample if that query proves to be a slow query.

To accomodate some of the queries produced by buck and logged to the log
file, we need to increase the size of the stack buffer that we use to
log a line.


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @dhruvsinghal and @Vaelatern to be potential reviewers.

wez:
Pushed as 38a2573

PR #333, opened 2016-08-15
Allow forcing the case sensitivity of a query

wez:
We default to respecting the case sensitivity of the watched root.
In some cases we are able to optimize a query if we know that the
tree is treated as case sensitive, so allow a query to force that
flag as appropriate.


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @dhruvsinghal and @sid0 to be potential reviewers.

wez:
Pushed as 7de84d2

PR #334, opened 2016-08-15
Add a cache for the parent path in the query context

wez:
A recent change made it O(depth-of-tree) to compute the wholename.
We already cache the wholename in the query executor, but there are
some operations where we are also interested in the name of the
parent of a file node.

This diff adds a simple cache for the most recently looked up file
parent and its full name string.


wez:
Pushed as c9a3c99ccce44f8327896e44d17268275a40a1de

PR #336, opened 2016-08-16
windows: suppress warning C4214

wez:
I recently added using bitfields to the string type, and the MS compiler
is complaining, let's see if this will keep it quiet

Refs: https://github.com/facebook/watchman/pull/325


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @dhruvsinghal and @otrempe to be potential reviewers.

PR #337, opened 2016-08-17
avoid crash if we sample a root concurrently with a recrawl

wez:
Since we call this function outside of a lock, there is a small window
of opportunity to read a NULL pointer for the watcher_ops member.


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @dhruvsinghal to be a potential reviewer.

PR #341, opened 2016-08-18
move state saving into its own thread

wez:
There are a couple of code paths that are made tricky because they may
deadlock around saving the state information.

This removes the source of the trickiness by having a dedicated thread
for saving the state; whenever we need to save the state, we signal the
thread to wakeup and it can block and take its sweet time without fear
of impacting the other threads that have more important things to do.

### DIFF
--- a/listener.c
+++ b/listener.c
@@ -756,6 +756,7 @@ bool w_start_listener(const char *path)
   }
 
   pthread_join(reaper_thread, &ignored);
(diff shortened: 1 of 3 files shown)


facebook-github-bot-6:

sid0:
LGTM

wez:
Pushed as 40d04bb

Issue #105, opened 2015-05-25
[Feature Request] Symlink support


JohnyDays:
Is there any performance / complexity reason for not supporting symlinks?

With the release of react-native, and the use of watchman in its packager, it's troubling that npm's (react-native's chosen package manager) endorsed method of developing local interdependent libraries with npm link is not supported, and having to commit to a remote repo and download on every change is obviously not ideal.

wez:
Watchman recursively watches files that are contained within a filesystem tree.
Symlinks can point to an arbitrary location on the filesystem.
Changes to symlink targets that are outside the tree are not observable.

Why not simply resolve the symlink and watch its target?  Because it isn't simple.  Here are a handful of reasons that make this a difficult prospect.  This is not an exhaustive list, it's just a few reasons I can rattle off the top of my head:

* The target of the symlink may not exist at the time we want to establish the watch, and neither may any of the parent directories in the symlink path.  The operating system will not allow us to watch a path that does not exist
* Any of the directory components in the target symlink path may themselves be a symlink and may change at an arbitrary time in their lifecycle in a way that is not detectable unless we watch every single directory component in the symlink target path
* The target of the symlink, or any of the directory components of the symlink target, may be a remote filesystem that doesn't support change notifications
* Each additional watched directory consumes more filesystem watch resources. There are finite limits on these, and not every system allows raising these limits.
* Because the hypothetical watches that we'd establish for each symlink are distinct from the watch used for the root of the real tree, we lose certain ordering guarantees when processing change notifications from the kernel.  We wouldn't be able to tell whether a change to the symlink target happened logically before or after changes that we observe in the main tree.  This important property is used to ensure that filesystem changes have been observed up to the logical point in time of a query.  More details on this can be found at https://facebook.github.io/watchman/docs/cookies.html
* Filesystem changes are fraught with TOCTOU (time of check, time of use) issues already.  Adding user-space resolution of symlinks increases the chances of more TOCTOU bugs dramatically; any component of the symlink path may change while we're processing information about them.

Correctly handling all of the above would be tremendously complex and still be error prone.  As a result, it is unlikely that we'll ever add support for resolving and tracking symlink targets.

wez:
Taking a step back from this, can you describe what you're trying to do and why symlinks are important?  I'm a casual and occasional node and npm user and am not familiar with `npm link`

JohnyDays:
I understand why this would be difficult, and I'll close the ticket and resign to using the above solution for developing local modules.

[...]

PR #347, opened 2016-08-23
Automatically watch symlink targets

AGFeldman:
Test Plan: Add unit tests

Differential Revision: https://reviews.facebook.net/D61509

### DIFF
--- a/root.c
+++ b/root.c
@@ -2020,6 +2020,7 @@ void watch_symlink_target(w_string_t *target) {
  * are themselves symlinks, this function gets called recursively called on
  * all the components of path. */
 static void watch_symlinks(w_string_t *path) {
+#ifndef _WIN32
   w_string_t *dir_name, *file_name;
   char link_target_path[WATCHMAN_NAME_MAX];
   ssize_t tlen = 0;
@@ -2070,6 +2071,7 @@ static void watch_symlinks(w_string_t *path) {
   if (file_name) {
     w_string_delref(file_name);
   }
+#endif
 }
 
 /** Process the list of observed changed symlinks and arrange to establish
--- a/root.c
+++ b/root.c
@@ -1967,7 +1967,7 @@ static w_string_t *get_normalized_target(w_string_t *target) {
 }
 
 // Requires target to be an absolute path
-static void watch_symlink_target(w_string_t *target) {
+static void watch_symlink_target(w_string_t *target, json_t *root_files) {
   char *watched_root = NULL, *relpath = NULL;
   w_string_t *normalized_target;
 
@@ -1988,27 +1988,18 @@ static void watch_symlink_target(w_string_t *target) {
     }
   } else {
     char *resolved, *errmsg;
-    json_t *root_files;
-    bool enforcing, res;
-    resolved = strdup(normalized_target->buf);
-    root_files = cfg_compute_root_files(&enforcing);
-    if (!root_files) {
-      w_log(W_LOG_ERR,
-            "watch_symlink_target: error computing root_files configuration "
-            "value, consult your log file at %s for more details\n", log_name);
+    bool res;
+    resolved = w_string_dup_buf(normalized_target);
+    res = find_project_root(root_files, resolved, &relpath);
+    if (!res) {
+      w_log(W_LOG_ERR, "No root project found to contain %s\n", resolved);
     } else {
-      res = find_project_root(root_files, resolved, &relpath);
-      if (!res) {
-        w_log(W_LOG_ERR, "No root project found to contain %s\n", resolved);
-      } else {
-        struct unlocked_watchman_root unlocked;
-        bool success = w_root_resolve(resolved, true, &errmsg, &unlocked);
-        if (!success) {
-          w_log(W_LOG_ERR, "watch_symlink_target: failed to watch %s\n",
-                resolved);
-        }
+      struct unlocked_watchman_root unlocked;
+      bool success = w_root_resolve(resolved, true, &errmsg, &unlocked);
+      if (!success) {
+        w_log(W_LOG_ERR, "watch_symlink_target: failed to watch %s\n",
+              resolved);
       }
-      json_decref(root_files);
     }
     // Freeing resolved also frees rel_path
     free(resolved);
@@ -2020,7 +2011,7 @@ static void watch_symlink_target(w_string_t *target) {
  * Since the target of a symbolic link might contain several components that
  * are themselves symlinks, this function gets called recursively called on
  * all the components of path. */
-static void watch_symlinks(w_string_t *path) {
+static void watch_symlinks(w_string_t *path, json_t *root_files) {
   w_string_t *dir_name, *file_name;
   char link_target_path[WATCHMAN_NAME_MAX];
   ssize_t tlen = 0;
@@ -2043,7 +2034,7 @@ static void watch_symlinks(w_string_t *path) {
     if (errno == EINVAL) {
       // The final component of path is not a symbolic link, but other
       // components in the path might be symbolic links
-      watch_symlinks(dir_name);
+      watch_symlinks(dir_name, root_files);
     } else {
       w_log(W_LOG_ERR,
           "readlink(%s) errno=%d tlen=%d\n", path->buf, errno, (int)tlen);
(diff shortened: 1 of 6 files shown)


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @wez, @anirbanr-fb and @dhruvsinghal to be potential reviewers.

ekryski:
@AGFeldman is this a resolution for #105? 😁

wez:
@ekryski not on its own, but it does bring us a step closer

facebook-github-bot-4:
@AGFeldman updated the pull request - [view changes](https://github.com/facebook/watchman/pull/347/files/4545e57ccfa3372a1b7b781be7a056f8789560ee..38370dd26ec6bf0bee6d8a062680a2b5e6a81e61)

facebook-github-bot-6:
@AGFeldman updated the pull request - [view changes](https://github.com/facebook/watchman/pull/347/files/38370dd26ec6bf0bee6d8a062680a2b5e6a81e61..17897c61097b3e2a4014b5179042ebfe855e0f3e)

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+    // We are already watching a root that contains this target
+    free(watched_root);
+    if (relpath) {
+      free(relpath);
+    }
+  } else {
+    char *resolved, *errmsg;
+    json_t *root_files;
+    bool enforcing, res;
+    resolved = strdup(normalized_target->buf);
```
wez: `w_string_dup_buf(normalized_target)`

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+    root_files = cfg_compute_root_files(&enforcing);
```
wez: would be great to factor this computation and the error check out to the caller of this function and just pass `root_files` down; there may be many thousands of symlinks in some repos, and on the first pass this call would be relatively hot/expensive

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+      if (!res) {
+        w_log(W_LOG_ERR, "No root project found to contain %s\n", resolved);
+      } else {
+        struct unlocked_watchman_root unlocked;
+        bool success = w_root_resolve(resolved, true, &errmsg, &unlocked);
+        if (!success) {
+          w_log(W_LOG_ERR, "watch_symlink_target: failed to watch %s\n",
+                resolved);
+        }
+      }
```
wez: don't forget to `w_root_delref(unlocked.root);` here in the success case, otherwise we'll leak the root.  (We need to add a nicer function to make this more obvious)

root.c
wez: (and by caller, I mean the function that iterates over the set of symlink candidates)

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+    }
+    // Freeing resolved also frees rel_path
+    free(resolved);
+  }
+  w_string_delref(normalized_target);
+}
+
+/** Given an absolute path, watch all symbolic links associated with the path.
+ * Since the target of a symbolic link might contain several components that
+ * are themselves symlinks, this function gets called recursively called on
```
wez: > called recursively called

has a redundant "called" :-)

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+  // We do not currently support symlinks on Windows, so comparing path to "/"
+  // is ok
+  if (!path || w_string_equal_cstring(path, "/")) {
+    return;
+  }
+  w_assert(w_is_path_absolute(path->buf),
+           "watch_symlinks: path %s is not absolute\n", path->buf);
+  dir_name = w_string_dirname(path);
+  file_name = w_string_basename(path);
+  tlen = readlink(path->buf, link_target_path, WATCHMAN_NAME_MAX);
```
wez: prefer `sizeof(link_target_path)` over `WATCHMAN_NAME_MAX` as this is less brittle in the face of future refactorings

wez: since `path` may be the result of a `w_string_dirname` call when we recurse (the `dir_name` assignment above), it may be a slice and therefore `path->buf` may not be nul-terminated.

You'll need to call `w_string_dup_buf(path)` to guarantee that you have a nul-terminated string.

root.c
```diff
@@ -2006,6 +2177,10 @@ static void io_thread(struct unlocked_watchman_root *unlocked)
     w_pending_coll_unlock(&unlocked->root->pending);
 
     if (!pinged && w_pending_coll_size(&pending) == 0) {
+#ifndef _WIN32
+      process_pending_symlink_targets(unlocked->root);
```
wez: prefer to pass `&unlocked` here instead, so that it is clear what the locking requirements are for `process_pending_symlink_targets`

tests/integration/test_symlink.py
```diff
@@ -124,3 +133,158 @@ def test_symlinkToFileDir(self):
...
+    @unittest.skipIf(os.name == 'nt', 'win')
+    def test_watchSymlinkTarget(self):
+        def make_roots():
+            rootA = self.makeRootAndConfig()
+            rootB = self.makeRootAndConfig()
+            rootC = self.makeRootAndConfig()
+            self.touchRelative(rootA, 'file')
+            self.touchRelative(rootB, 'file')
+            self.touchRelative(rootC, 'file')
+            os.symlink(os.path.join(rootC, 'file'), os.path.join(rootB, 'link'))
```
wez: Please add a comment something like `B/link -> C/file` so that it is quick and easy to understand the link structure at a glance :-)

tests/integration/test_symlink.py
```diff
@@ -124,3 +133,158 @@ def test_symlinkToFileDir(self):
...
+        self.assertFileList(root1prime, ['.watchmanconfig', 'file', 'link'])
+
+    @unittest.skipIf(os.name == 'nt', 'win')
+    def test_watchRelativeSymlinkTarget(self):
+        def make_roots():
+            rootA = self.makeRootAndConfig()
+            rootB = self.makeRootAndConfig()
+            self.touchRelative(rootA, 'file')
+            self.touchRelative(rootB, 'file')
+            os.symlink(os.path.relpath(os.path.join(rootB, 'file'),
```
wez: add a comment here to show an example of how the relpath looks in practice

root.c
```diff
@@ -1922,6 +1938,161 @@ static void notify_thread(struct unlocked_watchman_root *unlocked)
...
+    w_string_delref(dir_name);
+  }
+  if (file_name) {
+    w_string_delref(file_name);
+  }
+}
+
+/** Process the list of observed changed symlinks and arrange to establish
+ * watches for their new targets */
+static void process_pending_symlink_targets(w_root_t *root) {
```
wez: I suggested elsewhere that you pass in the unlocked pointer here.
It's also worth noting in a comment that it is safe for us to operate on `unlocked->root->pending_symlink_targets` because we only ever mutate it from the IO thread.

watchman.h
```diff
@@ -522,6 +522,9 @@ struct watchman_root {
    * of the suffix index.  Linkage via suffix_next */
   w_ht_t *suffixes;
 
+  /* collection of symlink targets that we try to watch */
+  struct watchman_pending_collection pending_symlink_targets;
```
wez: comment here to indicate that it is only safe to access this from the IO thread, either for reads or writes; it is not protected by the root lock.

facebook-github-bot-4:
@AGFeldman updated the pull request - [view changes](https://github.com/facebook/watchman/pull/347/files/17897c61097b3e2a4014b5179042ebfe855e0f3e..16c54745ec8a6e360b2f8ce1e32f835d2dca881b)

facebook-github-bot-3:
@AGFeldman updated the pull request - [view changes](https://github.com/facebook/watchman/pull/347/files/16c54745ec8a6e360b2f8ce1e32f835d2dca881b..965fd3fde82b25d5734242640f2e699a5e07f59c)

root.c
```diff
@@ -1999,6 +1999,7 @@ static void watch_symlink_target(w_string_t *target, json_t *root_files) {
       if (!success) {
         w_log(W_LOG_ERR, "watch_symlink_target: failed to watch %s\n",
               resolved);
+        w_root_delref(unlocked.root);
```
wez: this should be in the success case :-)

facebook-github-bot-7:
@AGFeldman updated the pull request - [view changes](https://github.com/facebook/watchman/pull/347/files/965fd3fde82b25d5734242640f2e699a5e07f59c..398b3605294065b8e22e56bdb97cde8a1371627f)

wez:
squashed and pushed as 6cf0bdd, thanks!

PR #348, opened 2016-08-24
Add load to parse a bser buffer from a file pointer

mzlee:
Step 1: refactor the header parsing into more reusable pieces
Step 2: Read into a local buffer and parse

### DIFF
--- a/python/pywatchman/bser.c
+++ b/python/pywatchman/bser.c
@@ -939,24 +939,15 @@ static PyObject *bser_loads_recursive(const char **ptr, const char *end,
   return NULL;
 }
 
-// This function parses the PDU header and provides info about the packet
-// Returns false if unsuccessful
-static int pdu_info_helper(PyObject *self, PyObject *args,
+static int _pdu_info_helper(const char *data, const char *end,
     uint32_t *bser_version_out, uint32_t *bser_capabilities_out,
-    int64_t *total_len_out) {
-  const char *start = NULL;
-  const char *data = NULL;
-  int datalen = 0;
-  const char *end;
+    int64_t *expected_len_out, off_t *position_out) {
   uint32_t bser_version;
   int64_t bser_capabilities = 0; // int64 because bunser_int requires it
   int64_t expected_len, total_len;
 
-  if (!PyArg_ParseTuple(args, "s#", &start, &datalen)) {
-    return 0;
-  }
-  data = start;
-  end = data + datalen;
+  const char *start;
+  start = data;
   // Validate the header and length
   if (memcmp(data, EMPTY_HEADER, 2) == 0) {
     bser_version = 1;
@@ -987,7 +978,34 @@ static int pdu_info_helper(PyObject *self, PyObject *args,
 
   *bser_version_out = bser_version;
   *bser_capabilities_out = (uint32_t) bser_capabilities;
-  *total_len_out = total_len;
+  *expected_len_out = expected_len;
+  *position_out = (off_t)(data - start);
+  return 1;
+}
+
+// This function parses the PDU header and provides info about the packet
+// Returns false if unsuccessful
+static int pdu_info_helper(PyObject *self, PyObject *args,
+    uint32_t *bser_version_out, uint32_t *bser_capabilities_out,
+    int64_t *total_len_out) {
+  const char *start = NULL;
+  const char *data = NULL;
+  int datalen = 0;
+  const char *end;
+  int64_t expected_len;
+  off_t position;
+
+  if (!PyArg_ParseTuple(args, "s#", &start, &datalen)) {
+    return 0;
+  }
+  data = start;
+  end = data + datalen;
+
+  if (!_pdu_info_helper(data, end, bser_version_out, bser_capabilities_out,
+                        &expected_len, &position)) {
+    return 0;
+  }
+  *total_len_out = (int64_t)(expected_len + position);
   return 1;
 }
 
@@ -1020,18 +1038,19 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
 {
   const char *data = NULL;
   int datalen = 0;
+  const char *start;
   const char *end;
   int64_t expected_len;
+  off_t position;
   PyObject *mutable_obj = NULL;
   const char *value_encoding = NULL;
   const char *value_errors = NULL;
   unser_ctx_t ctx = {1, 0};
-  int64_t bser_capabilities = 0; // int64 because bunser_int requires it
 
   static char *kw_list[] = {"buf", "mutable", "value_encoding", "value_errors",
                             NULL};
 
-  if (!PyArg_ParseTupleAndKeywords(args, kw, "s#|Ozz:loads", kw_list, &data,
+  if (!PyArg_ParseTupleAndKeywords(args, kw, "s#|Ozz:loads", kw_list, &start,
                                    &datalen, &mutable_obj, &value_encoding,
                                    &value_errors)) {
     return NULL;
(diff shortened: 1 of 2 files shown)


facebook-github-bot-4:
By analyzing the blame information on this pull request, we identified @dhruvsinghal, @sid0 and @Qwait to be potential reviewers.

facebook-github-bot-9:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/ac1d9cd45276076b0f9950aca864c90fbae8db46..6a825a66beef672b0ccb085d3e7ca2d5192acab1)

facebook-github-bot-1:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/6a825a66beef672b0ccb085d3e7ca2d5192acab1..e2dfd1dafdedf693fe9cca9a5c42c3815cc51a72)

facebook-github-bot-7:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/e2dfd1dafdedf693fe9cca9a5c42c3815cc51a72..b020b7b4909a27c5e9dc239df71774cdc796212c)

python/pywatchman/bser.c
```diff
@@ -1048,33 +1069,15 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+  data = start;
   end = data + datalen;
 
-  // Validate the header and length
-  if (memcmp(data, EMPTY_HEADER, 2) == 0) {
-    ctx.bser_version = 1;
-  } else if (memcmp(data, EMPTY_HEADER_V2, 2) == 0) {
-    ctx.bser_version = 2;
-  } else {
-    PyErr_SetString(PyExc_ValueError, "invalid bser header");
```
wez: the new version looks like it no longer raises this error; can you add it back?

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,87 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+  off_t position;
+  PyObject *fp_obj = NULL;
+  PyObject *mutable_obj = NULL;
+  const char *value_encoding = NULL;
+  const char *value_errors = NULL;
+  unser_ctx_t ctx = {1, 0};
+
+  PyObject *ret = NULL;
+
+  static char *kw_list[] = {"fp", "mutable", "value_encoding", "value_errors",
```
wez: I realize that this is just copy-pasted from above, but can this be made `const`?

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,87 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+    ctx.value_errors = value_errors;
+  }
+
+  if (!PyFile_Check(fp_obj)) {
+    PyErr_SetString(PyExc_ValueError, "invalid file object");
+    goto out;
+  }
+  fp = PyFile_AsFile(fp_obj);
+
+  rc = fread(header, 1, SNIFF_BUFFER_SIZE, fp);
```
wez: prefer `sizeof(header)` to `SNIFF_BUFFER_SIZE` here as that is less brittle in the face of a future refactor :-)

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,87 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+  if (rc != SNIFF_BUFFER_SIZE) {
+    PyErr_SetString(PyExc_ValueError, "bser data ended early");
+    goto out;
+  }
+  data = header;
+  if (!_pdu_info_helper(data, data + SNIFF_BUFFER_SIZE, &ctx.bser_version,
```
wez: lets make `data + SNIFF_BUFFER_SIZE` -> `data + rc`.  I know we're checking the length above, but it feels more future proof to make this respect the actually read length.

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,87 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+                        &ctx.bser_capabilities, &expected_len, &position)) {
+    goto out;
+  }
+  total_len = expected_len + position;
+  newdata = (char *)malloc(total_len);
+  if (newdata == NULL) {
+    goto out;
+  }
+  memcpy(newdata, header, SNIFF_BUFFER_SIZE);
```
wez: let's make the rest of these `SNIFF_BUFFER_SIZE`'s be `sizeof(data)` instead

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,87 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+  rc = fread(newdata + SNIFF_BUFFER_SIZE, 1, total_len - SNIFF_BUFFER_SIZE, fp);
+  if (rc != total_len - SNIFF_BUFFER_SIZE) {
+    PyErr_SetString(PyExc_ValueError, "bser data ended early");
```
wez: TBH, I have pretty mixed feelings about assuming that the read will get everything in one shot.  It's perfectly valid to have short reads depending on the type of the underlying file descriptor (eg; sockets tend to return smaller chunks)

Can you add a helper function called something like `fread_all` that keeps going for short reads and use that in place of the `fread` calls in here?

The meat of that function would be something like:

```
total_len = 0;
while (true) {
   len = fread(..., to_read);
   if (len <= 0) {
      return 0; // Error before we read everything
   }  
   total_len += len;
   to_read -= len;
   buf += len;
   if (to_read == 0) {
     return total_len;
  }
} 
```

wez:
Just some nits; looks good, thanks!

python/pywatchman/bser.c
mzlee: I agree that I don't think read will get everything in one shot, but my understanding is that fread is supposed to have that guarantee.  The only reason we would get less than the full amount is if there's an EOF or a read error

wez: fread makes no guarantee (http://pubs.opengroup.org/onlinepubs/009695399/functions/fread.html), it just happens to be the case that reading from a plain file has this greedy read behavior and that usually succeeds.

The FILE* may reference a socket or some other kind of descriptor where the reads tend to come in at around 4k, 8k or 16k unless someone bumped the tcp buffer sizes.

On windows and some flavors of BSD you can get ENOBUFS if you try to read more than the available kernel buffer space from a socket.  On Windows this can be as small as 8k, which leads to that read in the inner loop often looking like:

```
fread(...,MIN(to_read, 8192)...)
```

:-(

python/pywatchman/bser.c
mzlee: `_pdu_info_helper` takes care of the check and the throw for you.

python/pywatchman/bser.c
mzlee: ack

python/pywatchman/bser.c
mzlee: ack

facebook-github-bot-4:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/b020b7b4909a27c5e9dc239df71774cdc796212c..ed41dbcccae01a5b3eeacfbf05bfab3b7a4898b6)

facebook-github-bot-6:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/ed41dbcccae01a5b3eeacfbf05bfab3b7a4898b6..26b9c7bf3f4b5f494f32f28cad162b7a78d26455)

python/pywatchman/bser.c
```diff
@@ -1084,9 +1087,97 @@ static PyObject *bser_loads(PyObject *self, PyObject *args, PyObject *kw)
...
+    }
+    read_offset += rc;
+    read_remaining -= rc;
+  }
+  data = newdata + position;
+
+  ret = bser_loads_recursive(&data, end, &ctx);
+
+ out_free:
+  free(newdata);
```
wez: FWIW, `free(NULL)` is well defined and won't crash, so you can eliminate one of these labels; so long as newdata is initialized to NULL and not some random junk you can unconditionally jump down here and `free(newdata)`

facebook-github-bot-0:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/26b9c7bf3f4b5f494f32f28cad162b7a78d26455..20754e5107881d8feede84388c7176f30dc3cfcb)

python/pywatchman/bser.c
mzlee: ack and updated

facebook-github-bot-2:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/20754e5107881d8feede84388c7176f30dc3cfcb..d11752ac41a9cd9a6275cf4aeefab76e34bff2a2)

facebook-github-bot-7:
@mzlee updated the pull request - [view changes](https://github.com/facebook/watchman/pull/348/files/d11752ac41a9cd9a6275cf4aeefab76e34bff2a2..de538ac37d6c09d0f593bbb7df741eed60320d85)

PR #375, opened 2016-10-14
Whitelist index.js. This leaves out tests from npm installs

dawsonbotsford:
Read more about this feature here: https://docs.npmjs.com/files/package.json#files

### DIFF
--- a/node/bser/package.json
+++ b/node/bser/package.json
@@ -9,6 +9,9 @@
   "scripts": {
     "test": "node test/bser.js"
   },
+  "files": [
+    "index.js"
+  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/facebook/watchman"


dawsonbotsford:
The failed appveyor build shows a failure in a php file, but I changed only one line in the node implementation. For this reason, I'm led to believe that the build was broken before my changes were introduced.

wez:
Yeah, the integration tests can be a little flaky; I've restarted appveyor and booted travis.

Probably also a good idea to make the equivalent change in the bser package too?

https://github.com/facebook/watchman/blob/master/node/bser/package.json

wez:
... and looking again now I can see that that is exactly what this is, and that the other package.json already has this.   I think this PR is fine as-is; let's just give the CI a change to run and I'll land this and update the package in the npm registry.

dawsonbotsford:
Bump. Not sure what actions I could take here.

wez:
This is on me to push.  I've been a bit swamped but will take a look in the next few days

PR #380, opened 2016-10-30
w_string_suffix bugfix and unit test

kaitian521:
1. if a string ends with a '.', absolutely it dose not have suffix
2. `str->len - end >= sizeof(name_buf)` is more readable and comprehensible
3. add unit test for w_string_suffix.


wez:
Thanks for contributing!

Can you also add a unit test for this function?
We don't have one today, but it should be relatively easy to add one.

Take a look at `tests/argv.cpp` for an example; please add a `tests/string.cpp` file that demonstrates the expected behavior for a few different input strings, including this case where the name ends with a `.`.

You'll also need to edit `Makefile.am` and `winbuild/Makefile` to have the test build and run.

kaitian521:
i should have added some unit test for this function but i failed to find such files... so will i just add a unit test for this func, or i add unit tests for all functions in string.cpp and watchman_string.h? thank you for reply

wez:
Just coverage for the function you're changing is all that is needed here.  If you'd like to add coverage for other string functions, please send a separate pull request!

kaitian521:
@wez, I do not have windows in hand, so let ci run the test, and if something wrong, I will fix it

Makefile.am
```diff
@@ -268,6 +269,13 @@ tests_log_t_SOURCES = \
 	tests/log.cpp \
 	log.cpp
 
+tests_string_t_CPPFLAGS = $(THIRDPARTY_CPPFLAGS) @IRONMANCFLAGS@
+tests_string_t_LDADD = $(JSON_LIB) $(TAP_LIB)
+tests_string_t_SOURCES = \
```
wez: please add `tests/log_stub.cpp` in here; that will allow you to remove the `w_log` stub from the test case

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
+/* Copyright 2012-present Facebook, Inc.
+ * Licensed under the Apache License, Version 2.0 */
+
+#include "watchman.h"
+#include <algorithm>
+#include "thirdparty/tap.h"
+
+void w_log(int level, WATCHMAN_FMT_STRING(const char *fmt), ...) {
```
wez: please remove this function; there are stubs for this in `log_stub.cpp` that you should use instead

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+  unused_parameter(level);
+  unused_parameter(fmt);
+}
+
+void test_suffix() {
+
+  w_string_t *suffix = nullptr, *str = nullptr, *expect_suffix = nullptr;
+  str = w_string_new_typed("", W_STRING_UNICODE);
```
wez: prefer to use the newer smart pointer type for this:

```
w_string str("");
```

Note that it doesn't make too much sense to use `W_STRING_UNICODE` in here.  That is a tag to help the serialization layer recreate string or binary representations as appropriate.

Filenames in watchman are always binary which happens to be the default type when constructing a string via the `w_string` constructor, so you can just omit it.

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+  ok(!suffix, "end with dot");
+
+  str = w_string_new_typed("nosuffix", W_STRING_UNICODE);
+  suffix = w_string_suffix(str);
+  ok(!suffix, "no suffix");
+
+  str = w_string_new_typed(".beginwithdot", W_STRING_UNICODE);
+  suffix = w_string_suffix(str);
+  expect_suffix = w_string_new_typed("beginwithdot", W_STRING_UNICODE);
+  ok(w_string_equal(suffix, expect_suffix), "begin with dot");
```
wez: using `w_string` you can phrase this more clearly:

```
ok(w_string(".beginwithdot").suffix() == w_string("beginwithdot"), "begin with dot");
```

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+  ok(w_string_equal(suffix, expect_suffix), "java suffix");
+
+  // many '.' in name
+  str = w_string_new_typed("index.android.bundle", W_STRING_UNICODE);
+  suffix = w_string_suffix(str);
+  expect_suffix = w_string_new_typed("bundle", W_STRING_UNICODE);
+  ok(w_string_equal(suffix, expect_suffix), "multi dots suffix");
+
+  char too_long_name[129] = {0};
+  for (char &ch : too_long_name) {ch = 'a'; }
```
wez: prefer something like this:

```
std::string longName(129, 'a');
longName[0] = '.';
str = w_string(longName.data(), longName.size());
suffix = str.suffix();
```

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+  too_long_name[0] = '.';
+  str = w_string_new_len_typed(too_long_name, sizeof(too_long_name), W_STRING_UNICODE);
+  suffix = w_string_suffix(str);
+  ok(!suffix, "too long suffix");
+
+  if (str) w_string_delref(str);
```
wez: you won't need these delrefs when you switch to `w_string`.

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+  if (suffix) w_string_delref(suffix);
+  if (expect_suffix) w_string_delref(expect_suffix);
+
+}
+
+int main(int argc, char **argv)
```
wez: since we're in c++, you can just omit the argument names of unused parameters:

```
int main(int, char**) {
```

tests/string.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+  (void)argc;
```
wez: and then you won't need these

winbuild/Makefile
```diff
@@ -136,7 +136,8 @@ TEST_DIR_SRCS_CPP=\
 	tests\ht_test.cpp \
 	tests\ignore_test.cpp \
 	tests\log_stub.cpp \
-	tests\log.cpp
+	tests\log.cpp \
+	tests\string.cpp
 
 OBJS=$(SRCS_CPP:.cpp=.obj) $(SRCS_C:.c=.obj)
 TEST_OBJS=$(TEST_SRCS_CPP:.cpp=.obj) $(TEST_SRCS_C:.c=.obj)
```
wez: github won't let me comment on precisely the correct line, but you'll need to add `tests\string.exe` to the list of `TESTS=` that is a couple of lines below this in order for the windows build to build and run your new test

wez:
FWIW, https://github.com/facebook/watchman/commit/7a9d768db0d964a60b34f649b82b4e8610e81f87 adds a string.cpp test case and the associated build plumbing, so you'll need to rebase your changes; the good news is that you can focus on adding the suffix test cases to that now existing file rather than worrying about making it build!

kaitian521:
i see the commit you pick up above which save me a lot of time.

kaitian521:
maybe I can rename all files that have suffix '.', so I think this issue can be closed.

Makefile.am
```diff
@@ -134,8 +134,8 @@ TESTS = \
 		tests/bser.t \
 		tests/ignore.t \
 		tests/pending.t \
-		tests/string.t \
 		tests/log.t \
+		tests/string.t \
```
wez: please revert the changes in this file; since you're now adding to the existing test, this pull request doesn't need to make changes here :-)

Makefile.am
```diff
@@ -259,6 +259,7 @@ tests_log_t_SOURCES = \
 	tests/log.cpp \
 	log.cpp
 
+
```
wez: same here

string.cpp
```diff
@@ -624,6 +624,11 @@ w_string_t *w_string_suffix(w_string_t *str)
       }
       *buf = '\0';
       return w_string_new_typed(name_buf, str->type);
+    } else {
```
wez: prefer to phrase this block as:

```
} else if (str->len - end >= sizeof(name_buf)) {
   // We haven't found the '.' yet but the suffix will never fit in our local buffer
   return nullptr;
}
```

kaitian521:
sorry for my poor english  😊

PR #393, opened 2016-11-30
Fix BSER for nodejs

### LINKED ISSUE #392: [nodejs] Using unix timestamp for since does not work
When I try to issue a subscribe query to watchman from nodejs with the since value being a number watchman throws an error.

```
Error: failed to parse query: invalid value for 'since'
    at BunserBuf.<anonymous> (node_modules/fb-watchman/index.js:95:23)
    at emitOne (events.js:96:13)
    at BunserBuf.emit (events.js:188:7)
    at BunserBuf.process (node_modules/bser/index.js:289:10)
    at node_modules/bser/index.js:244:12
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

The arguments for the subscribe query is:
```
{ 
  expression: [ 'allof', [ 'name', ["...", "long", "array", "of", "files", "..."], 'wholename' ] ],
  fields: [ 'name', 'mtime_ms', 'exists' ],
  since: 1480513073
 }
```

According to the doc and as far as I can see in [clockspec.cpp](https://github.com/facebook/watchman/blob/9619ff88d3504938e079a38c7778504cf6b66d06/clockspec.cpp#L29..L33) it should be allowed to send an integer.

Am I doing something wrong?

danez:
This fixes several issues with the nodejs implementation of BSER. Javascript is tricky as it treats all numbers as float, so there are some drawbacks of this implementation.

* Fixed the MAX_INT16 and MAX_INT32 constants to be correct. They were both +1 which led to `argument is out of bounds` if trying to dump `32768` or `2147483648` as int
* Fixed detection of integers inside `dump_any()` by checking if it is an integer with the polyfill outline in this [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill).
  * This has the drawback that a float cannot be enforced anymore, as `1.0` will also be detected as int because `1.0 === 1`
  * The only real solution to properly fix all cases which I can see, is to have a `new Double()` class or `new Integer()`class, which a user would need to supply to enforce the type.
* Changed `readInt()` when reading an 64bit integer to check if the value can be handled by node.js. If yes it returns the primitive number otherwise returns the instance of `Int64`. This imho is nice, even if the return value in some cases could be either `number` or `Int64` (not sure what is the actual limit)
  * This is a breaking change for users who always expect a `number > MAX_INT32` to be an instance of Int64. (e.g. `mtime_ms` in watchman was always a `Int64` and could now be a `number`)
  * If not desired, let me know, I can revert this.
* Changed the tests to use deepStrictEqual to ensure primitive types are checked with `===`

Fixes #392

### DIFF
--- a/node/bser/index.js
+++ b/node/bser/index.js
@@ -120,6 +120,9 @@ Accumulator.prototype.peekInt = function(size) {
 
 Accumulator.prototype.readInt = function(bytes) {
   var ival = this.peekInt(bytes);
+  if (ival instanceof Int64 && isFinite(ival.valueOf())) {
+    ival = ival.valueOf();
+  }
   this.readOffset += bytes;
   return ival;
 }
@@ -207,8 +210,8 @@ var ST_NEED_PDU = 0; // Need to read and decode PDU length
 var ST_FILL_PDU = 1; // Know the length, need to read whole content
 
 var MAX_INT8 = 127;
-var MAX_INT16 = 32768;
-var MAX_INT32 = 2147483648;
+var MAX_INT16 = 32767;
+var MAX_INT32 = 2147483647;
 
 function BunserBuf() {
   EE.call(this);
@@ -490,8 +493,13 @@ function dump_int(buf, val) {
 function dump_any(buf, val) {
   switch (typeof(val)) {
     case 'number':
-      buf.writeByte(BSER_REAL);
-      buf.writeDouble(val);
+      // check if it is an integer or a float
+      if (isFinite(val) && Math.floor(val) === val) {
+        dump_int(buf, val);
+      } else {
+        buf.writeByte(BSER_REAL);
+        buf.writeDouble(val);
+      }
       return;
     case 'string':
       buf.writeByte(BSER_STRING);
--- a/node/bser/test/bser.js
+++ b/node/bser/test/bser.js
@@ -10,7 +10,7 @@ var template =  "\x00\x01\x03\x28" +
                 "\x1e\x0c\x03\x19" ;
 
 var val = bser.loadFromBuffer(template);
-assert.deepEqual(val, [
+assert.deepStrictEqual(val, [
   {"name": "fred", "age": 20},
   {"name": "pete", "age": 30},
   {"age": 25}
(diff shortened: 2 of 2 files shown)


wez:
Wow, this is little embarrassing... thanks for taking this on!

I think I'm fine with the idea of this making a breaking API change so long as it doesn't bite anyone in practice.  Floats a relatively rare in the FB usage of watchman and BSER, so I don't anticipate that that will hurt any of the commonly used tools here.

cc: @mostafaeweda for Nuclide and js-fu on the FB side.  I wonder if this might explain some of that weird BSER errors you mentioned a while back?  Do you anticipate that this will break Nuclide?  Would you mind also reviewing the JS here; clearly my implementation was off in some important details :-p

wez:
(CI may be a bit flaky at the moment because of some pretty major changes in the core that haven't been checked out on all platforms)

node/bser/index.js
```diff
@@ -120,6 +120,9 @@ Accumulator.prototype.peekInt = function(size) {
 
 Accumulator.prototype.readInt = function(bytes) {
   var ival = this.peekInt(bytes);
+  if (ival instanceof Int64 && isFinite(ival.valueOf())) {
```
mostafaeweda: when is is expected to not be finite?
In that case, the return value here can either be `number` or `Int64`
I'd rather check the `isFinite` on the consumer side of this function's return value.

node/bser/index.js
```diff
@@ -120,6 +120,9 @@ Accumulator.prototype.peekInt = function(size) {
...
+    ival = ival.valueOf();
+  }
   this.readOffset += bytes;
   return ival;
```
mostafaeweda: let's have a consistent return result - either `number` or `Int64`.

mostafaeweda:
Nuclide sits at the top level, far away from `bser` - https://github.com/facebook/nuclide/blob/master/pkg/nuclide-watchman-helpers/lib/WatchmanClient.js#L284-L286 expecting request / response objects
So, I highly doubt this would break it, but it may indeed fix the `bser` deserialization errors we had every now and then.

node/bser/index.js
danez: As I wrote in the other comment, it already returns both depending on the size of the integer.

danez:
Thanks, would be nice if you could also make a new release. :)

wez:
Yeah, we're just getting some things squared away; also need to release the python bindings shortly as well

PR #394, opened 2016-12-02
fix resource leak

iskander-saitbatalov:
Fix resource leak at the correct operation of the program (forget to close file).
---------------
Iskander Saitbatalov


iskander-saitbatalov:
Ohh. Sorry for bad pull-request. I will change it and try another one.

PR #395, opened 2016-12-02
fix work with file in test

iskander-saitbatalov:
1. fix segfault if file doesn't exist, you can check this if you will delete file or change permission to load
2. fix resource leak (close file)


wez:
Why does that file not exist for you?  It's part of the repo and will always be there.  It's fine for the test to fail by crashing if the test file is not present

wez:
Closing due to lack of feedback

PR #399, opened 2016-12-08
Forced bser module import by using pdu_info

zphricz:
Because demandimport lazily import bser, this fallback approach doesn't actually seem to work. I ran into issues with it when trying to get watchman working on windows when I didn't have the compiled bser module.

### DIFF
--- a/python/pywatchman/__init__.py
+++ b/python/pywatchman/__init__.py
@@ -42,6 +42,9 @@ import time
 # so fall back to a pure Python implementation.
 try:
     from . import bser
+    # Demandimport causes modules to be loaded lazily. Force the load now
+    # so that we can fall back on pybser if bser doesn't exist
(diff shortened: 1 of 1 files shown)


bhamiltoncx:
The `bser` module contains a shim which automatically uses the Python version if the C version isn't available.

Maybe that's not working on Windows?

I'm not sure what the goal of this no-op statement is.

sid0:
> I'm not sure what the goal of this no-op statement is.

The goal of it is to force a module load. `demandimport` disables module loading until the first time a property is accessed on the module, and it does just that.

(FB-internal link) see https://fburl.com/578220911

wez:
The CI failures are due to some flakiness that is very slightly better in master as of a bit earlier today (rebasing should help), but don't block this diff; it looks good to me.

PR #412, opened 2017-01-05
Don't swallow exceptions in WatchmanConnection.

mkillianey:
Summary:
During deserialization, the SerializedStreamMessageSupplier was swallowing exceptions
and returning null.  It was doing this to adhere to the com.google.common.base.Supplier
contract.  Instead, use the java.util.concurrent.Callable interface, which is (essentially)
functionally identical, but has a signature that supports checked exceptions.

Since the supplier is running in an executor, it is essential that the executor *not* swallow all exceptions, since that makes it immune to canceling by InterruptedExceptions.


JonShemitz:
I don't see how this fixes the core-pegging bug. If the socket breaks, you need to open a new socket to the new process. Where is that code?

wez:
Also summoning @bhamiltoncx as a potentially interested party from the Buck side of things

mkillianey:
@JonShemitz is discussing a problem found in code that uses this watchman Java library.

Q:  Why does this fix a core-pegging bug?
A:  Because the Supplier was swallowing exceptions and returning null, if the deserializer got in a bad state where it was perpetually failing, the WatchmanConnection would get stuck in the while loop here...

https://github.com/mkillianey/watchman/blob/2a2ac89d1c35a273a43325098676d5870a465c2e/java/src/com/facebook/watchman/WatchmanConnection.java#L238-L265

...where it would perpetually throw/swallow exceptions inside what is essentially a while(true) loop.

Q:  Who is responsible for opening a new socket?
A:  Probably the user of this library, since they passed in the socket to the WatchmanClientImpl constructor.  (There isn't a no-arg constructor.)

It might be a good future improvement to add a utility to this library that detects when the socket dies (for example, if someone bounces the watchman process), but for right now, the behavior is just to die.  This change just makes it die without hogging resources.

java/src/com/facebook/watchman/WatchmanClientImpl.java
```diff
@@ -71,7 +72,7 @@ public Boolean get() {
 
   @VisibleForTesting
   WatchmanClientImpl(
-      Supplier<Map<String, Object>> inputSupplier,
+      Callable<Map<String, Object>> inputSupplier,
```
styurin: `inputSupplier` -> `inputCallable`

java/test/com/facebook/watchman/WatchmanClientTest.java
```diff
@@ -53,7 +53,7 @@
   public void setClient() throws IOException {
     withWatchProject = true;
     mClient = new WatchmanClientImpl(
-        mInputSupplier,
+      mInputStream,
```
styurin: indentation is off

java/test/com/facebook/watchman/WatchmanTestBase.java
```diff
@@ -22,31 +22,26 @@
...
-import com.google.common.base.Supplier;
 import org.junit.Assert;
 import org.junit.Before;
 
 public class WatchmanTestBase {
 
   protected ByteArrayOutputStream mOutputStream;
   protected BlockingQueue<Map<String, Object>> mObjectQueue;
-  protected Supplier<Map<String, Object>> mInputSupplier;
+  protected Callable<Map<String, Object>> mInputStream;
```
styurin: why stream?

java/src/com/facebook/watchman/WatchmanConnection.java
```diff
@@ -45,7 +45,7 @@
 
   private final ListeningExecutorService outgoingMessageExecutor;
   private final ExecutorService incomingMessageExecutor;
-  private final Supplier<Map<String, Object>> inputMessageSupplier;
+  private final Callable<Map<String, Object>> inputMessageCallable;
```
styurin: It looks like this rename hides the meaning of this variable. This is something that provides input messages. The new name looks like it's something that is called on an input message or something similar, but not the original purpose.

styurin:
These changes looks good, I have a few minor notes about naming.

java/BUCK
```diff
@@ -8,8 +8,8 @@ java_binary(
 java_library(
   name = 'watchman-lib',
   srcs = glob(['src/**/*.java']),
-  source = '1.6',
-  target = '1.6',
+  source = '7',
```
mkillianey: Sorry to add more changes to this, but in generating an IJ project using `buck project`, I realized that the source level had been incorrectly set to 6, whereas the code is this library is actually written to 7.

java/src/com/facebook/watchman/WatchmanClientImpl.java
```diff
@@ -71,12 +72,12 @@ public Boolean get() {
 
   @VisibleForTesting
   WatchmanClientImpl(
-      Supplier<Map<String, Object>> inputSupplier,
-      OutputStream outputStream,
+      Callable<Map<String, Object>> incomingMessageGetter,
```
mkillianey: @styurin , I know you didn't like the name inputCallable or inputMessageCallable (and I agree with you), so I'm trying out this, instead:  `incomingMessageGetter` for the Callable that gives you messages from watchman, and `outgoingMessageStream` for the OutputStream that sends serialized messages to wachman.  Better?  Or do you have some other suggestion?

java/test/com/facebook/watchman/WatchmanConnectionTest.java
```diff
@@ -295,5 +297,83 @@ public void onFailure(Throwable t) {
       entry++;
     }
   }
+
+  @Test
+  public void propagatesDeserializationException() throws Exception {
```
mkillianey: Added this unit test to make sure that if we see the BserEofException that's been showing up in stack traces, that the exception will get propagated back to the user of this library.

java/test/com/facebook/watchman/WatchmanConnectionTest.java
```diff
@@ -295,5 +297,83 @@ public void onFailure(Throwable t) {
...
+    try {
+      listenableFuture.get(5, TimeUnit.SECONDS);
+      Assert.fail("Should not be able to get a result from an interrupted operation");
+    } catch (InterruptedException e) {
+      // This is reasonable behavior...
+      Assert.assertSame("If InterruptedException, should be the one thrown by thread",
+          expectedException.get(), e);
+    } catch (ExecutionException e) {
+      // ...and this is also reasonable behavior.
+      Assert.assertSame("If ExecutionException, cause should be exception that interrupted thread",
```
mkillianey: This double-catch is weird, but reading the documentation for the interface method `Future#get(int,TimeUnit)`, it seems to be up to the implementation's discretion whether it throws a bare `InterruptedException` or wraps it in an `ExecutionException`.  

Since both are arguably reasonable interpretations (and since the most likely user case is for the caller to generate a stack trace or print the message, in which case these are equally useful), as long as it's one or the other, that seems reasonable to me.

java/test/com/facebook/watchman/WatchmanConnectionTest.java
```diff
@@ -295,5 +297,83 @@ public void onFailure(Throwable t) {
...
+    final Semaphore exitedCallable = new Semaphore(0);
+    final AtomicReference<Thread> threadRef = new AtomicReference<>();
+    final AtomicReference<InterruptedException> expectedException = new AtomicReference<>();
+    Callable<Map<String, Object>> incomingMessageGetter = new Callable<Map<String, Object>>() {
+      @Override
+      public Map<String, Object> call() throws Exception {
+        try {
+          threadRef.set(Thread.currentThread());
+          enteredCallable.release();
+          Thread.sleep(1);
```
styurin: Can we avoid sleeping in a test? Maybe use a latch to make sure thread is at a specific location and then interrupt it? Sleeping is usually associated with non-determinism and flakiness in tests.

java/BUCK
```diff
@@ -32,14 +32,15 @@ java_library(
...
-  target = '1.7',
+  source = '7',
+  target = '7',
   deps = [
     ':watchman-lib',
     '//third-party/guava:guava',
     '//third-party/hamcrest:hamcrest-2',
     '//third-party/junit:junit',
     '//third-party/mockito:mockito',
+    '//third-party/objenesis:objenesis',
```
styurin: is this necessary?

java/src/com/facebook/watchman/WatchmanClientImpl.java
styurin: Yeah, this is better

asp2insp:
LGTM

PR #417, opened 2017-01-08
Use pageid instead of id and use kramdown instead of redcarpet

JoelMarcey:
`pageid` is the method to get the `id` of a page
kramdown is the default markdown processor for GitHub pages

https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

The pageid change is necessary to prevent not being able to display the sidebar items on the docs page.

screenshot 2017-01-07 17 40 27

vs. 

screenshot 2017-01-07 21 29 25


sid0:
Thanks!

wez:
Thanks @JoelMarcey!

Issue #232, opened 2016-03-15
watchman-make: add --run option as alternative to --make


sebm:
I use watchman-make to automatically trigger a script on filesystem changes, which I've accomplished in a fairly hacky way:

> `watchman-make -p '**/*.js' '**/*.css' --make my_script.sh -t ''`

The `-t ''` is necessary because otherwise I get the error

> # No targets were specified, nothing to do.

Would you consider allowing an empty target or at least a `--no-target` flag for this kind of use case? I could even write it myself, I just want to make sure it conforms to your design standards.

wez:
It works this way because it is intended to be used primarily with a build tool that you can instruct to make (potentially) multiple different targets.

I'm amenable to making a simpler UI for non-make use cases, how about something like:

```
watchman-make -p '**/*.js' '**/*.css' --run my_script.sh
```

that will just run `my_script.sh` when any of the matching patterns change.

Does that sound good to you?  If so, would you like to submit a pull request to implement that?

sebm:
Hey @wez, that would work for me. I'm amenable to making a pull request when I get a chance…watch this space.

wez:
Looking forward to it, thanks!

joseluisq:
@wez `--run` option is a great idea.
But, Is there any alternative way to do this with the current release?
Thanks.

wez:
@joseluisq the original post shows exactly how to do this today

joseluisq:
> @joseluisq the original post shows exactly how to do this today

Yeah works thanks!

TDAbboud:
I got around to implementing this feature and submitted a PR. I would appreciate any feedback.

joseluisq:
@TDAbboud awesome!

wez:
This will be included in the 4.8 release which should happen in the next week or so

sebm:
Thank you @TDAbboud !

PR #425, opened 2017-01-19
add a --run option to execute a given script

TDAbboud:
This option provides an alternative to the --make
flag. You can now specify a pattern and provide
the path to a script to run upon changes being detected.
Ex:
    `watchman-make -p '*.c' '*.h' --run my_script.sh`

Implements #232

### DIFF
--- a/python/bin/watchman-make
+++ b/python/bin/watchman-make
@@ -16,27 +16,20 @@ def patterns_to_terms(pats):
         terms.append(['match', p, 'wholename', {'includedotfiles': True}])
     return terms
 
-
 class Target(object):
-    """ Represents a makefile target that we'd like to build
+    """ Base Class for a Target
 
     We track the patterns that we consider to be the dependencies for
     this target and establish a subscription for them.
 
     When we receive notifications for that subscription, we know that
-    we should execute the command to build this target.
+    we should execute the command.
     """
-
-    def __init__(self, name, make, targets, patterns):
+    def __init__(self, name, patterns, cmd):
         self.name = name
-        self.make = make
-        self.targets = targets
         self.patterns = patterns
+        self.cmd = cmd
         self.triggered = False
-
-    def __repr__(self):
-        return '{make=%r targets=%r pat=%r}' % (
-            self.make, self.targets, self.patterns)
 
     def start(self, client, root):
         query = {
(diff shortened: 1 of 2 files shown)


website/_docs/watchman-make.md
```diff
@@ -89,3 +89,19 @@ $ watchman-make -p '*.c' '*.h' 'Makefile*' 'tests/**/*.py' 'tests/**/*.c' -t all
 this will execute `make all integration` if you change any top level `*.c` file
 *or* test source file.
 
+### Run Scripts
+
```
wez: Insert:

```
*Since 4.8.*
```

here so that folks know which version this will appear in.  We're prepping to tag this right now; just have a couple of things to tidy up, so it should be out by the end of next week, assuming things go to plan.

python/bin/watchman-make
```diff
@@ -16,27 +16,20 @@ def patterns_to_terms(pats):
...
-        self.make = make
-        self.targets = targets
         self.patterns = patterns
-        self.triggered = False
-
-    def __repr__(self):
-        return '{make=%r targets=%r pat=%r}' % (
-            self.make, self.targets, self.patterns)
+        self.cmd = cmd
+        self.triggered = triggered
```
TDAbboud: I am assuming you are referring to both `self.make` and `self.targets`. They were only needed to construct the command that was going to run, but now the subclass will set the `self.cmd` which will run. I will set `self.triggered = false` and then remove it from the argument list.

python/bin/watchman-make
```diff
@@ -68,13 +61,36 @@ class Target(object):
...
+
+class MakefileTarget(Target):
+    """ Represents a Makefile target that we'd like to build. """
+    def __init__(self, name, make, targets, patterns):
+        self.make = make
+        self.targets = targets
         cmd = "%s %s" % (self.make, " ".join(self.targets))
-        print('# Execute: `%s`' % cmd, file=sys.stderr)
-        subprocess.call(cmd, shell=True)
+        super(MakefileTarget, self).__init__(name, patterns, cmd, triggered=False)
```
TDAbboud: You are right. As per the comment above I will just remove this argument.

python/bin/watchman-make
```diff
@@ -68,13 +61,36 @@ class Target(object):
...
+
+    def __repr__(self):
+        return '{make=%r targets=%r pat=%r}' % (
+            self.make, self.targets, self.patterns)
+
+class RunTarget(Target):
+    """ Represents a script that we'd like to run. """
+    def __init__(self, name, runfile, patterns):
+        self.runfile = runfile
+        cmd = "%s" % self.runfile
```
TDAbboud: Good catch!

TDAbboud:
I have fixed the issues you mentioned and re-pushed the branch. I ran a few tests as you mentioned and you can see the output below. One interesting note is that you can also execute a target-centric build using the `--run 'make integration'` syntax (see the last image below).

Original `make watch` using the target-centric build:
make_watch

Run the run_integrations.sh script, which executes the `make integration` build
run_integrations_script

Run the target-centric build using the `--run` syntax
run_make_watch

PR #435, opened 2017-03-02
Make `bser_load()` match Python implementation for optional args

bhamiltoncx:
The C implementation of `bser.load()` makes all arguments required, but the Python implementation  only requires the first argument.

This makes the C implementation match the Python implementation.


bhamiltoncx:
Just included a test.

PR #451, opened 2017-04-24
Add troubleshooting hint when watching many files

naehrstoff:
The problem our whole team ran into was that watchman tried to watch the node_modules folder as well. This exceeded the file limit in macOS Sierra and was solved by ignoring that folder through a config file.

Also fixes broken link to “remove the warning” section.

### DIFF
--- a/website/_docs/troubleshooting.markdown
+++ b/website/_docs/troubleshooting.markdown
@@ -179,8 +179,19 @@ limits for the number of files.
 
 ### How do I resolve this?
 
-[Follow these directions](
-/watchman/docs/troubleshooting.html#i-39-ve-changed-my-limits-how-can-i-clear-the-error)
+If the problem is due to trying to watch too many files (this can happen easily in
+a project with a `node_modules` folder, for example), you should instruct watchman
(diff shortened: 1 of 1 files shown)


wez:
Thanks for contributing!  Regarding the error that you saw, can you expand on that?
Watchman doesn't keep many files open and shouldn't cause you trip over this sort of a limit by itself, so the implication is that there is either a bug in watchman, or that some other process is keeping a lot of files open based on its interaction with watchman.

I'd like to understand that situation a little better, and then we can tweak the advice from the troubleshooting section accordingly.  Ignoring the dir is effective at this sort of thing, but only if you know that you never want to check in the associated files in source control, and that none of your tools ever want to use watchman to know when those files have changed.  We'd need to at least mention this caveat before making a blanket suggestion to the user.

wez:
@naehrstoff: would you mind responding to my comment above when you have the chance?  Thanks!

naehrstoff:
Yes, I'm sorry I haven't replied yet – I wanted to send you some more information that takes some time to get at. I'll try to answer soon!

wez:
I'm going to close this out for now; when you have time to provide the feedback requested above, we can re-open and resume the discussion.  Thanks!

PR #452, opened 2017-04-27
Provide human readable errors for log commands

cpick:
Previously, erroneous or the wrong number of arguments for `log` or
`log-level` commands would throw and return a cryptic error.

Validate log-level command's arguments server-side and return human
readable errors.


wez:
Thanks for including tests!

wez:
Pushed with a couple of tweaks to make our internal linters happy; thanks!

Issue #444, opened 2017-04-03
Windows: Endless loop in InMemoryView::ioThread


otrempe:
Hi!

I am currently struggling debugging the watchman process being caught in an endless loop within InMemoryView::ioThread.

I could not identify the exact cause of the problem and reproduce it systematically.

I generally observe it while compiling a large Visual Studio solution for which the output directory is under a watched root. If I add the output directory to "ignore_dirs", it seems to be just worse.

I also observed it while doing some rebases with mercurial.

I tried to debug what's going on. I was first suprised to see that files under my ignore_dirs entry were still stat-ed. Why?

Then I realized the files were endlessly "stolen" from the pending collection ( coll->stealItems(); ) and re-added to the collection in the last "if" of "InMemoryView::statPath". So I guess this is why the pending collection is never entirely processed, hence the endless loop.

I thought it might be related to the presence of symlinks in the tree, but real files are also caught in the this loop.

I am willing to put some effort to fix this, but I would need some help to start in the right direction  given my limited background playing in the watchman source code.

This is a major showstopper for me to roll out watchman to 70+ Windows users.

wez:
I think I might have a fix for this sitting in my tree.  I need to get this cleaned up a bit before I can land it.  I'll see if I can get it in better shape to at least share with you.

otrempe:
That's great news. We can run watchman with this custom patch for a few days and if it works properly, I can turn it into a pull request.

otrempe:
Hi @wez, any progress on this?

If you don't plan to land the patch soon, you could send me a draft, so I can validate it fixes the problem. And maybe I could clean it on my side...

wez:
I'm (finally!) looking at windows stuff this week; let's see whether we can sort this out at the same time

wez:
OK, those changes should now all be in the most recent build, and this link should always resolve to the most recent build if you just want to try it out quickly without building it yourself:

https://ci.appveyor.com/api/projects/wez/watchman/artifacts/watchman.zip?branch=master&job=Environment:+WATCHMAN_WIN7_COMPAT%3D

otrempe:
Awesome! I played with it this afternoon and it seems to work perfectly. I deploy it immediately to a few developers. If we don't have issues within a few days, we will roll it out to the whole team. I expect some very happy co-workers next week!

wez:
How's this working out for you?  Can we close this issue?

otrempe:
[...]

PR #455, opened 2017-04-29
Fix a Windows 7 compatibility issue in WinDirHandle::readDir

epatenaude:
On Windows 7, in WinDirHandle::readDir(), a system_error would be thrown
when calling GetFileInformationByHandleEx with FileFullDirectoryInfo.
FileFullDirectoryInfo is not supported before Windows 8
(https://msdn.microsoft.com/en-us/library/windows/desktop/aa364228(v=vs.85).aspx).
To fix this, we use an implementation based on the FindFirstFile and
FindNextFile functions.

Before c2dec83 (refactor w_dir_open, nov 2016), readDir() would simply
return NULL on Win7 which hid the problem.

### DIFF
--- a/winbuild/dir.cpp
+++ b/winbuild/dir.cpp
@@ -209,7 +209,11 @@ class WinDirHandle : public watchman_dir_handle {
       return readDirWin8();
     }
     catch (const std::system_error& err) {
-      // Fallback on Win7 implementation
+      if (err.code().value() != ERROR_INVALID_PARAMETER) {
+        throw;
+      }
+      // Fallback on Win7 implementation. FileFullDirectoryInfo
+      // parameter is not supported before Win8
       win7_ = true;
       return readDirWin7();
     }
--- a/winbuild/dir.cpp
+++ b/winbuild/dir.cpp
@@ -9,19 +9,21 @@ using watchman::Win32Handle;
 
 namespace {
 class WinDirHandle : public watchman_dir_handle {
+  std::wstring dirWPath_;
   Win32Handle h_;
   FILE_FULL_DIR_INFO* info_{nullptr};
   char __declspec(align(8)) buf_[64 * 1024];
+  HANDLE hDirFind_{nullptr};
   char nameBuf_[WATCHMAN_NAME_MAX];
   struct watchman_dir_ent ent_;
 
  public:
   explicit WinDirHandle(const char* path, bool strict) {
     int err = 0;
-    auto wpath = w_string_piece(path).asWideUNC();
+    dirWPath_ = w_string_piece(path).asWideUNC();
 
     h_ = Win32Handle(intptr_t(CreateFileW(
-        wpath.c_str(),
+        dirWPath_.c_str(),
         GENERIC_READ,
         FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
         nullptr,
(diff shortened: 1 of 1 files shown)


winbuild/dir.cpp
```diff
@@ -48,22 +50,104 @@ class WinDirHandle : public watchman_dir_handle {
...
 
   const watchman_dir_ent* readDir() override {
-    if (!info_) {
-      if (!GetFileInformationByHandleEx(
-              (HANDLE)h_.handle(), FileFullDirectoryInfo, buf_, sizeof(buf_))) {
-        if (GetLastError() == ERROR_NO_MORE_FILES) {
-          return nullptr;
+    if ((getenv("WATCHMAN_WIN7_COMPAT") && 
+         getenv("WATCHMAN_WIN7_COMPAT")[0] == '1')) {
+      // Before Windows 8, the FileFullDirectoryInfo parameter is not supported when
```
wez: This seems like an incomplete thought being expressed in the comment; can you fix it up, or just truncate it to "not supported"?

winbuild/dir.cpp
```diff
@@ -48,22 +50,104 @@ class WinDirHandle : public watchman_dir_handle {
   }
 
   const watchman_dir_ent* readDir() override {
```
wez: would you mind breaking this up a little bit so that the win7 and win8+ versions are separate functions?
That will help keep the diff smaller; logically you'd be renaming the existing method and adding the helper from this comment, and another method for the win7 logic.

```
const watchman_dir_ent* readDir() override {
   if (win7_) {
      return readDirWin7();
   }
   return readDirWin8();
}
```

winbuild/dir.cpp
```diff
@@ -48,22 +50,104 @@ class WinDirHandle : public watchman_dir_handle {
...
-    if (!info_) {
-      if (!GetFileInformationByHandleEx(
-              (HANDLE)h_.handle(), FileFullDirectoryInfo, buf_, sizeof(buf_))) {
-        if (GetLastError() == ERROR_NO_MORE_FILES) {
-          return nullptr;
+    if ((getenv("WATCHMAN_WIN7_COMPAT") &&
```
wez: Please put this detection logic in the constructor and set a `bool win7_` member variable.
It's also not super clear how this is detected at runtime without these env vars being set.  The env vars are used elsewhere to force this code path in the tests, but we also need a way to catch this and switch on `win7_` when these env vars are not set.

One approach for this is to try the win8 function, then if we get an error code that tells us it isn't implemented, set the win7 flag and try the other method.

epatenaude:
Thank you for the review! As suggested, I restructured the code to isolate the two versions and added detection of the error to automatically fallback on the win7 version. 

Regarding the failing tests on the AppVeyor CI, let me know if you feel it could be related to my last commit, I will look at them. Thanks!

winbuild/dir.cpp
```diff
@@ -91,23 +62,158 @@ class WinDirHandle : public watchman_dir_handle {
...
+  }
+
+  const watchman_dir_ent* readDir() override {
+    if (win7_) {
+      return readDirWin7();
+    }
+    try {
+      return readDirWin8();
+    }
+    catch (const std::system_error& err) {
```
wez: Sorry for the lag in responding; I was traveling at the end of last week and am just catching up.

Can we check for a specific error code for this?  The concern I have here is that a directory could have been removed since the time we decided that we need to look at it and the time we get here.   That would mean that we'd do twice as much work for a removed dir here on windows 8.

That may not sound bad, but for a very large tree or sub-tree that has been recursively deleted, this could translate into quite a lot of pressure on the filesystem layers of the system.

Is there a definitive error code that you can check for here before unilaterally falling back to win7?

I'd suggest logging the `err.code()` to figure out what actually happens on win7 and then updating this to match.

Thanks!

epatenaude: Thanks for the feeback, Your are absolutely right, the fallback should be scoped, if possible, only to our case, Otherwise we could end up using both implementations on all platforms without even knowing it.

wez:
Thanks for this!  I re-ordered a couple of hunks to make the diff a bit smaller and easier to review and pushed it

epatenaude:
Great! Thank you for the feedback and your time! Our 75+ Windows users (25 Win7, 25 Win8.1 and 25 Win10) will certainly appreciate the performance improvement of the status with our 125K files Hg repo. However, as reported by @otrempe in https://github.com/facebook/watchman/issues/444, we are still waiting for a fix for this issue before we can deploy it. If you have a few minutes to look at it, it would be greatly appreciated. Thanks again!

PR #459, opened 2017-05-03
[WIP] add rust support

sid0:
This isn't complete yet -- just putting this up for initial review for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,8 +75,10 @@ m4/
 .libs/
 *.lo
 *.la
-tests/integration/cppclient.t
+/tests/integration/cppclient.t
 libtool
 ltmain.sh
 cppclient/watchmanclient.pc
 /website/.jekyll-metadata
+/rust/**/Cargo.lock
+/rust/**/target/
--- a/rust/serde_bser/Cargo.toml
+++ b/rust/serde_bser/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "serde_bser"
+version = "0.1.0"
+authors = ["Siddharth Agarwal <sid0@fb.com>"]
+
+[dependencies]
+byteorder = "1.0"
+bytes = "0.4"
+error-chain = "*"
+serde = "1"
+serde_bytes = "0.10"
+serde_derive = "1"
--- a/rust/serde_bser/src/de/bunser.rs
+++ b/rust/serde_bser/src/de/bunser.rs
@@ -1,0 +1,137 @@
+//! Internal stateless code for handling BSER deserialization.
+
+use byteorder::{ByteOrder, NativeEndian};
+
+use de::read::{DeRead, Reference};
+use errors::*;
+use header::*;
+
+#[derive(Debug)]
+pub struct Bunser<R> {
+    read: R,
+    scratch: Vec<u8>,
+}
+
+pub struct PduInfo {
+    pub bser_capabilities: u32,
+    len: i64,
+    start: i64,
+}
+
+impl<'de, R> Bunser<R>
+    where R: DeRead<'de>
+{
+    pub fn new(read: R) -> Self {
+        Bunser {
+            read,
+            scratch: Vec::with_capacity(128),
+        }
+    }
+
+    /// Read the PDU off the stream. This should be called in the beginning.
+    pub fn read_pdu(&mut self) -> Result<PduInfo> {
+        {
+            let magic = self.read_bytes(2)?;
+            if magic.get_ref() != &EMPTY_HEADER[..2] {
+                bail!("invalid magic header {:?}", magic);
+            }
+        }
+        let bser_capabilities = self.read.next_u32(&mut self.scratch)?;
+        let len = self.check_next_int()?;
+        let start = self.read_count();
+        Ok(PduInfo {
+               bser_capabilities,
+               len,
+               start,
+           })
+    }
+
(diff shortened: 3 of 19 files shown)


rust/serde_bser/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "serde_bser"
+version = "0.1.0"
+authors = ["Siddharth Agarwal <sid0@fb.com>"]
+
+[dependencies]
+byteorder = "1.0"
+bytes = "0.4"
+error-chain = "*"
```
jsgf: Using `"*"` versions is generally bad form for Cargo.toml. crates.io won't let you upload anything with a `"*"` version constraint in it.

rust/serde_bser/src/header.rs
```diff
@@ -0,0 +1,43 @@
+//! Header constants for BSER.
+
+pub const EMPTY_HEADER: &[u8] = b"\x00\x02\x00\x00\x00\x00\x05\x00\x00\x00\x00";
```
jsgf: would using an array literal be clearer here?
```
pub const EMPTY_HEADER: &[u8] = &[0, 2, 0, 0, 0, 0, 5, 0, 0, 0, 0]; // or 0x00, etc
```

rust/serde_bser/src/de/bunser.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    #[inline]
+    pub fn discard(&mut self) {
+        self.read.discard();
+    }
+
+    /// Return a borrowed or copied version of the next n bytes.
+    #[inline]
+    pub fn read_bytes<'s>(&'s mut self, len: i64) -> Result<Reference<'de, 's, [u8]>> {
+        // TODO: what if the length can't be converted to a usize?
```
jsgf: It's probably reasonable to panic; it would imply either len is negative, or usize is 32 bits and len is > 2^32.

rust/serde_bser/src/ser/test.rs
```diff
@@ -0,0 +1,56 @@
...
+const BASIC_SERIALIZED: &[u8] = b"\x00\x02\x00\x00\x00\x00\x04\xcb\x00\x01\x03\x04\r\x03\t\
+                                  test_list\x00\x03\x05\x03\x03\x03\x04\x03*\x04\xdb\x03\x05\x00\
+                                  \x00\x08\x00\r\x03\ntest_tuple\x00\x03\x02\r\x03\x03foo\x07\
+                                  \x18-DT\xfb!\t@\r\x03\ttest_enum\r\x03\x08TestUnit\r\x03\x0e\
+                                  test_enum_list\x00\x03\x03\x01\x03\x01\r\x03\x0bTestNewtype\
+                                  \x02\x03\tBSER test\x01\x03\x01\r\x03\tTestTuple\x00\x03\x02\x03*\
+                                  \x06\xff\xff\xff\xff\xff\xff\xff\x7f\x01\x03\x01\r\x03\n\
+                                  TestStruct\x00\x03\x02\r\x03\x03abc\n\r\x03\x03\
+                                  def\r\x03\x04\xf0\x9f\x92\xa9";
+
```
jsgf: Are these tests dependent on the endianness of the test machine?

rust/serde_bser/src/de/bunser.rs
wez: panic for bad input data seems a bit harsh; ideally would bubble up as some kind of InvalidData error?

rust/serde_bser/src/ser/test.rs
wez: Yeah, bser is local endianness dependent so that it is valid to reference data directly out of the byte stream.  I don't know if anyone has run any of our other tests on a big endian box; if they have, they haven't reported a test failure to us :)

rust/serde_bser/src/de/bunser.rs
wez: oh, misread this.  Yeah, panic should be ok in this case

.gitignore
```diff
@@ -75,8 +75,10 @@ m4/
 .libs/
 *.lo
 *.la
-tests/integration/cppclient.t
+/tests/integration/cppclient.t
 libtool
 ltmain.sh
 cppclient/watchmanclient.pc
 /website/.jekyll-metadata
+/rust/**/Cargo.lock
```
stuhood: Not sure if this is intended to build reproducibly, but: it's a good idea to commit `Cargo.lock` files.

Happy to see this!

sid0: The general rule is to commit Cargo.lock files for binaries but not for libraries.

stuhood: Gotcha. But presumably there is a "root" somewhere in here that you're planning to load? A `cdylib` or something? I believe that you'd want a reproducible build starting from whatever your root is.

sid0: If I understood you correctly, you're asking for what happens when we decide to link the Rust bindings into the Watchman binary. (Please correct me if I'm wrong! :) )

We're not planning to load the Rust bindings into the main Watchman binary at all, actually. The Rust bindings are going to be provided for Rust projects to use to talk to Watchman out of process. I agree that if and when we decide to load the Rust bindings into the main Watchman binary, we'd want a Cargo.lock.

Hope that helps :)

stuhood: Ahh, I see. Yep, that helps. Thanks!

PR #471, opened 2017-05-20
Speculative fix for appveyor build issue

wez:


wez:
I've pushed this internally, should sync out here soonish

PR #472, opened 2017-05-20
[travis] switch to trusty beta

sid0:
Separate this out from #468 because it might be causing test failures.


sid0:
The failures look like they're unrelated (they happen on macOS too, which shouldn't be affected by this change).

PR #476, opened 2017-06-02
WIP: gcc 4.8 on travis miscompares std::error_code stuff!?

wez:
This is looking like a compiler/environment specific nuance.  Trying to add some debugging as my local version of gcc 4.8 doesn't manifest the same problem.

### DIFF
--- a/error_category.cpp
+++ b/error_category.cpp
@@ -51,6 +51,14 @@ const std::error_category& inotify_category() {
 
 bool error_category::equivalent(const std::error_code& code, int condition)
     const noexcept {
+  log(DBG,
+      "error_category::equivalent? code={",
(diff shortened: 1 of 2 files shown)


sid0:
Could we just say you need gcc 5 or above to compile?

wez:
I don't think we can constrain the version; 4.8 on my devserver didn't exhibit the same problem.  It looks like the issue was with the implicit conversion from `std::error_condition` to `std::error_code` for the `std::errc` enum values.  The explicit `make_error_code` calls seem to resolve it, so I'll just land that portion of this PR.

sid0:
Ah. that's an unfortunate bug.

PR #478, opened 2017-06-03
turn off strict symlink check in OpenFileHandleOptions.openDir().

zonr:
Use OpenFileHandleOptions.strictOpenDir() to enable the strict symlink
rules explicitly.

Before this change, `watchman::openFileHandle()` always enforces strict
symlink rules because `strictNameChecks` is true even in
`OpenFileHandleOptions.openDir()`.

### DIFF
--- a/FileSystem.h
+++ b/FileSystem.h
@@ -63,6 +63,7 @@ struct OpenFileHandleOptions {
   static inline OpenFileHandleOptions openDir() {
     OpenFileHandleOptions opts;
     opts.readContents = 1;
+    opts.strictNameChecks = false;
     opts.followSymlinks = 1;
     return opts;
   }


wez:
It's on by default by design.  Are you running into a problem with this?  Can you provide more context on the issue you're having?

zonr:
The PR is somewhat unrelated to the problem I encountered but happened to address the issue. Here's the story. I built my watchman with `./configure --disable-statedir` on macOS. With this option, watchman will setup its state dir under `TMPDIR`:

```cpp
 733 static void setup_sock_name(void)
 734 {
 735   const char *user = compute_user_name();
 736 
 737   watchman_tmp_dir = get_env_with_fallback("TMPDIR", "TMP", "/tmp");
```

On my macOS, `TMPDIR` looks like:

```bash
watchman (master) $ printenv TMPDIR
/var/folders/4y/qb6vbn5j6m1dd_f1vswbcvqc0000gn/T/
```

`/var` on macOS is a symbolic link to `/private/var`:

```bash
$ ls -1l /var
lrwxr-xr-x@ 1 root  wheel  11 12 10 21:13 /var -> private/var
```

This makes the strict symlink check in `watchman::openFileHandle()` unhappy and aborts the program with the following (uncaught) exception:

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: open(/var/folders/4y/qb6vbn5j6m1dd_f1vswbcvqc0000gn/T//zonr-state): opened path doesn't match canonical path /private/var/folders/4y/qb6vbn5j6m1dd_f1vswbcvqc0000gn/T/zonr-state: No such file or directory
```

Even if `/var` is itself not a symlink, the state dir looks like (notice the _double slashes_) :

```
/var/folders/4y/qb6vbn5j6m1dd_f1vswbcvqc0000gn/T//zonr-state
```

Strict symlink check would still fail because the exact match with canonical path is required.

After studying the code a while, I noticed that `watchman::openFileHandle()` called by `compute_file_name()` is not intended to perform the strict symlink checks as stated in comment:

```cpp
 541 static void compute_file_name(char **strp,
 542     const char *user,
 543     const char *suffix,
 544     const char *what)
 545 {
         ...
 585       auto dirp =
 586           w_dir_open(state_dir, false /* don't need strict symlink rules */);
 587 
```

However, because `OpenFileHandleOptions::strictNameChecks` is initialized to `1`, it makes no difference between `OpenFileHandleOptions::openDir()` and `OpenFileHandleOptions::strictOpenDir()`. The strict rules are always enforced.

I think that initializing `strictNameChecks` to `0` would make code more intuitive. If strict symlink check is desired, we could send another patch to pass `true /* strict */` in `compute_file_name()`'s `w_dir_open` (or even better, omit it because `strict` is an default argument of value `true` in  `w_dir_open`. See `watchman_opendir.h`.)

wez:
We need the strict checks to correctly detect filename case changes and avoid problems with symlinks, so we can't take this PR as-is; you can see that it breaks the integration tests on platforms with case insensitive filesystems.

I think what you really want is this diff:

```
diff --git a/fbcode/watchman/FileSystem.h b/fbcode/watchman/FileSystem.h
--- a/fbcode/watchman/FileSystem.h
+++ b/fbcode/watchman/FileSystem.h
@@ -63,6 +63,7 @@
   static inline OpenFileHandleOptions openDir() {
     OpenFileHandleOptions opts;
     opts.readContents = 1;
+    opts.strictNameChecks = false;
     opts.followSymlinks = 1;
     return opts;
   }
```

which turns off the checks for the non-strict opendir.  I think this was an oversight on my part, based on this being missing here and being redundantly turned on in the strict flavor of the options method. Can you try this out and let me know if that resolves the issue for you?

zonr:
Thanks! It's my bad to neglect the break in Windows build. Thanks for pointing out. Your suggested diff works well for me. Another way would be turn on strict checks in `queryFileInfo()`. What do you think?

```
diff --git a/FileSystem.h b/FileSystem.h
index 377249ef..c937bf5f 100644
--- a/FileSystem.h
+++ b/FileSystem.h
@@ -52,11 +52,12 @@ struct OpenFileHandleOptions {
   OpenFileHandleOptions()
       : followSymlinks(0), closeOnExec(1), metaDataOnly(0), readContents(0),
         writeContents(0), create(0), exclusiveCreate(0), truncate(0),
-        strictNameChecks(1), caseSensitive(CaseSensitivity::Unknown) {}
+        strictNameChecks(0), caseSensitive(CaseSensitivity::Unknown) {}
 
   static inline OpenFileHandleOptions queryFileInfo() {
     OpenFileHandleOptions opts;
     opts.metaDataOnly = 1;
+    opts.strictNameChecks = true;
     return opts;
   }
 
```

zonr:
Rebase.

wez:
This looks good; thanks for reporting it and thanks for taking a stab at fixing it!  I've imported it and once it's run through some more checks on our internal CI, I'll land it.

Thanks!

Issue #496, opened 2017-06-28
Build fails on Fedora 25 with the error: redhat-hardened-cc1: No such file or directory


amcghie:
Build of watchman fails on Fedora 25 with the error:
```
gcc: error: /usr/lib/rpm/redhat/redhat-hardened-cc1: No such file or directory
error: command 'gcc' failed with exit status 1
Makefile:3849: recipe for target 'py-build' failed
make[1]: *** [py-build] Error 1
make[1]: Leaving directory '/home/andrewmcghie/dev/watchman'
Makefile:867: recipe for target 'all' failed
make: *** [all] Error 2
```
It appears that on fedora (and probably other Redhat derivatives) you also need to install the following package: ``redhat-rpm-config``.  Documentation update required.

PR #497, opened 2017-06-28
Updated documentation to include the installation of redhat-rpm-config

amcghie:
As per issue #496 installation on Fedora requires the ``redhat-rpm-config`` package.


wez:
That sounds like a really weird dependency for python.  What specifically does that rpm install that makes the build succeed?

Are you actually building an rpm, or are you building watchman by hand?

sid0:
(sending it back to you @amcghie for more questions)

wez:
I'm going to just close this out for now; if/when you do have time to revisit, please re-open it!

PR #502, opened 2017-07-24
Use a wheel format when publishing the watchman pypi package

orf:
The Python wheel format is the preferred way of distributing Python packages with C extensions. This just adds the `bdist_wheel` step to the build script.

The `wheel` package needs to be installed on any development machine this is being run on.

### DIFF
--- a/python/publish-pypi.sh
+++ b/python/publish-pypi.sh
@@ -20,4 +20,4 @@
 #    password: <pypi_password>
 #
 
-python setup.py sdist upload -r pypi
+python setup.py sdist bdist_wheel upload -r pypi


sid0:
Looks great, thanks.

I guess that whenever we ship a new version of pywatchman, we'll need to build and upload it separately on every platform. Some CI integration for this would be nice, but I don't think it's blocking.

orf:
Unfortunately yep. You would want to just run the 'bdist_wheel' command and
not the sdist one as well.

To be honest it's only Windows and Mac that need it the most, in that order.

PR #504, opened 2017-07-25
Make watchman path configurable in Python client

indygreg:
Not all installs may have `watchman` on PATH. Some clients may
wish to point pywatchman to a custom binary.

This commit makes the watchman binary path configurable. It still
defaults to "watchman." Backwards compatibility is maintained.

The file appears to use inconsistent naming styles for arguments
and class members. I wasn't sure what naming style to use, so I just
picked one.


sid0:
Thanks Greg!

Issue #506, opened 2017-07-26
Provide some documentation on how to compile watchman on Windows


jlecomte:
This issue is related to issue #494. I would like to distribute the watchman binary along with my desktop application on Windows and Mac, which the license allows me to do. On Windows, it seems like additional DLLs are needed, which I can also distribute. However, I was wondering whether I could compile watchman with fewer dependencies so that I don't have to distribute multiple binaries for it to work on a windows system. I tried to compile watchman using cygwin, and I started getting a bunch of errors (sorry, I'm a Mac guy, I don't have a ton of experience with Windows or Cygwin) A little bit of documentation on how to compile watchman on Windows would go a long way.

jlecomte:
After installing the latest version of Visual Studio Community edition on Windows 7 (Parallels VM), I cloned the watchman repo and changed the following line inside `make.bat`:

```
c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat
```

to:

```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat
```

Then, I ran `make.bat` and managed to compile `watchman.exe`. Progress!

jlecomte:
Changing `/MD` to `/MT` at https://github.com/facebook/watchman/blob/master/winbuild/Makefile#L8 allowed me to build watchman with the VC runtime libs statically linked.

wez:
Thanks for digging in!  I think it's reasonable to make `/MT` the default for watchman; would you like to submit a pull request for just that change?

wez:
the community edition build change is also something like to see us handle, but it feels like a separate issue and the main thing you're interesting in resolving is the dynamic deps :)

Issue #494, opened 2017-06-23
I get MSVCP140.dll error when testing if watchman is working


erlendureyv:
Hi,

I followed the instructions on http://facebook.github.io/watchman/docs/install.html for Windows but I keep getting MSVCP140.dll error when testing if watchman is working by running watchman -v via command prompt (with administrative privileges).  I am using Windows 8.1 Version 6.3 Build 9600 according to winver via command prompt.

Regards, Erlendur

![image](https://user-images.githubusercontent.com/29660165/27486380-eed4f900-581f-11e7-9d98-1547e856cbd9.png)

wez:
Can you try the advice from this page?

https://answers.microsoft.com/en-us/windows/forum/windows_10-performance/msvcp140dll-is-missing-in-my-win-10/1c65d6b0-68b8-4b59-b720-3e6a33038389

wez:
@erlendureyv: did the comment above help you resolve this issue?

erlendureyv:
Hi,
I have not had the time to test it.  Should be able to do so in the next few days.
Regards, Erlendur


Sent from my Samsung Galaxy smartphone.
-------- Original message --------From: Wez Furlong <notifications@github.com> Date: 03/07/2017  19:42  (GMT-05:00) To: facebook/watchman <watchman@noreply.github.com> Cc: erlendureyv <erlendur.eyvindarson@gmail.com>, Mention <mention@noreply.github.com> Subject: Re: [facebook/watchman] I get MSVCP140.dll error when testing if
  watchman is working (#494) 
@erlendureyv: did the comment above help you resolve this issue?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




[...]

PR #513, opened 2017-08-04
Statically link the run-time library

jlecomte:
Trivial PR to statically link the Windows run-time library as a follow up to issues #494 and #506.

cc @wez


wez:
Thanks!

wez:
damn, looks like our sync machinery attributed the commit to me instead of you; that doesn't normally happen.  I'll follow up with the folks that own that piece to see what went wrong

PR #526, opened 2017-09-06
Migrate to CircleCI 2.0

serima:
### Summary

Bumped up CircleCI 2.0 to improve the build time.

### Questions

I have passed the test without installing PHP, why is explicit installation of PHP in the old config file?
Does the version specification of each language specified in the old config file have a deep meaning?


.circleci/config.yml
```diff
@@ -0,0 +1,17 @@
...
+     steps:
+       - checkout
+       - run:
+           name: Install dependencies
+           command: apt-get update && apt-get install libtool m4 automake npm -y
+       - run:
+           name: System Information
+           command: |
+             python -V
+             node -v
```
wez: this is currently failing the build :-(

.circleci/config.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  build:
+     docker:
+      - image: python:2.7
+     working_directory: ~/repo
+     steps:
+       - checkout
+       - run:
+           name: Install dependencies
+           command: apt-get update && apt-get install libtool m4 automake npm -y
```
wez: we used to handle this sort of thing in `travis/deps.sh`, but when they switched to their new container setup, we had to move the package installation info to `.travis.yml`.   It would be good to change the `command:` here to something like:

```
   command: apt-get update && ./travis/deps.sh
```

and in `travis/deps.sh`:

```
case `uname` in
   Linux)
      apt-get install <INSERT LIST OF PACKAGES FROM TRAVIS.YML HERE>
```

That would allow folks to run `travis/deps.sh` when they want to make sure the pre-requisites are all installed, even when they are not on travis.

.circleci/config.yml
serima: Sorry, I will fix it soon.

serima:
@wez Thank you for your review and explaining background.

> How much of a build time improvement are we seeing with these changes?

CI passed by hotfix. 10 min -> 7 min.

Latest build on facebook/watchman: https://circleci.com/gh/facebook/watchman/1033
This PR's build: https://circleci.com/gh/serima/watchman/16

I will continue to improve to:
- try to run our tests with Python 3.x concurrently
- try to fork `travis/deps.sh` to install dependencies for Circle

serima:
@wez Both versions passed the tests. Since it is being tested in parallel, there is no major impact on build time.

Python 2.7 : https://circleci.com/gh/serima/watchman/22
Python 3.6 : https://circleci.com/gh/serima/watchman/23

We also need to do this:

> If you have implemented workflows on a branch in your GitHub repository, but the status check never completes, there may be status settings in GitHub that you need to reset.

https://circleci.com/docs/2.0/workflows-waiting-status/

wez:
Ahh, except that it broke travis:

```
0.05s$ ./travis/deps.sh
+case `uname` in
++uname
+apt-get install libtool m4 automake npm -y
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

Let's go back on my suggestion to put the install step into `travis/run.sh` and just put it back in the circle defs where you had it originally, and then I'll merge this in.

Thanks!

serima:
Sorry for not confirming the result of Travis.
Modified not to use `./travis/deps.sh`. Please check it.

PR #527, opened 2017-09-06
Fix watchman on solaris/illumos.

niclasr:
Fix watchman on solaris/illumos.
Review wanted it might not be the most optimal fixes.
Tested on Openindiana Hipster.
I have tested the commands "watch", "find" and "since" and they work without a hassle.


stream_unix.cpp
```diff
@@ -74,20 +77,32 @@ class UnixStream : public watchman_stream {
...
+    socklen_t len = sizeof(cred);
     credvalid =
         getsockopt(fd.fd(), SOL_LOCAL, LOCAL_PEERCRED, &cred, &len) == 0;
+#elif defined(SO_RECVUCRED)
+    cred = nullptr;
+    credvalid = getpeerucred(fd.fd(), &cred) == 0;
 #endif
   }
 
+#ifdef SO_RECVUCRED
```
wez: ... and then we can remove this bit.

stream_unix.cpp
```diff
@@ -163,6 +178,11 @@ class UnixStream : public watchman_stream {
     if (cred.cr_uid == getuid() || cred.cr_uid == 0) {
       return true;
     }
+#elif defined(SO_RECVUCRED)
+    uid_t ucreduid = ucred_getruid(cred);
```
wez: the `unique_ptr` will mean that we use `cred.get()` here to pass the raw pointer value on.

main.cpp
```diff
@@ -93,7 +95,10 @@ static bool lock_pidfile(void) {
     return false;
   }
 
-  dprintf(fd.fd(), "%d", mypid);
+  std::ostringstream oss;
```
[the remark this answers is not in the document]
niclasr: nice.

niclasr:
Here is a gist with output from make integration.
https://gist.github.com/niclasr/f231b87d44eaa591b5812ea6d083f1b7
It looks like a disaster, most of the errors are "connection refused" and "WatchmanError('EOF on CLI process transport')" so there is still stuff to be fixed.

wez:
Yeah, the EOF thing sounds like a crash/uncaught exception.

That last line of output: `Preserving output in /tmp/watchmantestF2v_AA` shows you where the logs from the run were stored.  What I typically do to run down the problems there is:

```
less /tmp/watchmantest*/inst*/log
```

Then press `G` to jump to the bottom and look for a stack trace; if I see one, then I go dig into the relevant portion of the code.  Otherwise, `:n` to go to the log file from the next instance and repeat.

You can also do:

```
rm -rf /tmp/watchmantest* ; ./runtests.py --concurrency=1 --keep-if-fail
```

to (clean up any test data from a prior run and) reduce the number of concurrent watchman instances run by the tests; that can be helpful in some situations.

If for some reason the backtrace isn't useful (depends on how good the `backtrace()` implementation is on Solaris), you can run this instead:

```
rm -rf /tmp/watchmantest* ; ./runtests.py --debug-watchman --keep-if-fail
```

and the test runner will print out the pid of the watchman instance and wait for you to attach to it with your favourite debugger before starting the tests so that you can use it to examine the stack trace.

stream_unix.cpp
```diff
@@ -74,20 +77,32 @@ class UnixStream : public watchman_stream {
   struct ucred cred;
 #elif defined(LOCAL_PEERCRED)
   struct xucred cred;
+#elif defined(SO_RECVUCRED)
+  ucred_t* cred;
```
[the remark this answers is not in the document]
niclasr: That didn't work out of the box had to create a function object for it to work.

 ``` 
  struct ucred_deleter {
    void operator()(ucred_t *utp) {ucred_free(utp);}
  };
  std::unique_ptr<ucred_t, ucred_deleter> cred;
```

File: log.cpp [hunk 5]
```diff
@@ -235,13 +235,13 @@ static void crash_handler(int signo, siginfo_t* si, void*) {
   }
 
   if (si) {
```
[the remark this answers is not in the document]
niclasr: I suggest `#if defined(__sun) && !defined(HAVE_DPRINTF)` and adding dprintf to ac_check_funcs just in case solaris gets dprintf. It is after all a posix.1-2008 function.

niclasr:
Thanks for sharing ways to look at the logs.
Indeed there was a backtrace.
The failing log is here:
https://gist.github.com/niclasr/eaa81f53020507c3a7d1cee739ffb0bf
The backtrace demangled is here:
https://gist.github.com/niclasr/1534246b9ab39757fed23a47b5c06c6f
some speculative analysis:
The crashing line seems to be line 542 in string.cpp
tracing the data back leads to pending->path
pending comes from the collection.
I belive that the line "w_string_delref(name);"
in portfs.cpp is causing the problem.
Integration output from fixing this and adding the changes 
you suggested:
https://gist.github.com/niclasr/c35fd3e00645a0fa91942c2c53445e4b
It is test_trigger.TestTriggerLocalBser2.test_legacyTrigger that fails here
and test_subscribe.TestSubscribeLocalBser2.test_immediate_subscribe
is not there.
I have run integration without the trigger test.
https://gist.github.com/niclasr/ff49dcc2b2fe421f279518cdf26fad6c
Here test_subscribe.TestSubscribeLocalBser2.test_immediate_subscribe
passes so I belive that I have fixed the particular problem that caused 
the backtrace.
I have no idea what causes test_trigger.TestTriggerLocalBser2.test_legacyTrigger
to fail, there is no backtrace.
Running top when the test is run show that watchman goes down but doesn't
come back up. 

I have looked at the code where startWatchDir and startWatchFile is used.
watchman_watcher.h says that startWatchDir should return null on error.
but the code where it is used only checks for exceptions. Is the comment wrong?
I want to know what the return api is so I can write the methods correctly.

watcher/portfs.cpp
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  return true;
+}
+
+bool PortFSWatcher::startWatchFile(struct watchman_file* file) {
+  w_string name;
+  bool success = false;
+
+  name = w_dir_path_cat_str(file->parent, file->getName());
```
wez: re: your comment about the delref, I'd make this line: `auto name = ...`
and then remove the explicit delref later on; `w_string` (the return type from that function) will correct destruct and free things as appropriate.  Likewise, if there are other `w_string_delref` calls in this file, you can and should make a similar change

wez:
re: comments on those functions, they're probably a bit stale since we flipped things over to c++, sorry!   I believe that we've standardized on those throwing exceptions on error there, as we can provide more context to the caller about what failed.

re: this line in your test output:

```
File "/export/home/niclas/tmp/watchman-niclasr2/tests/integration/WatchmanInstance.py", line 209, in _waitForSuspend
    status = out.splitlines()[-1]
```

this bit of code may be sensitive to the formatting and options in the `ps` executable; there's already a `SunOS` test happening in the code and it's probably not quite right:

https://github.com/facebook/watchman/blob/master/tests/integration/WatchmanInstance.py#L196-L215

The intent is to suspend all the threads of the watchman server, then make some changes, then resume the watchman server and check that we didn't miss anything.  That helper method is used to test whether the process has suspend and resumed correctly so that we are appropriately synchronized.

niclasr:
The change to _waitForSuspend fixes test_trigger.TestTriggerLocalBser2.test_legacyTrigger.
I have run integration with the latest changes:
https://gist.github.com/niclasr/71982f3bb842e777a8f3515a818537bc
In this output some php tests fail and mention dev. Dev is mentioned in configure.ac
as a reason to use 64-bit builds on Solaris. I tried to do a 32-bit build and run the integration test on that.
https://gist.github.com/niclasr/5faa98558e2bf9359c2a37dd7fa69dc6
In this all php tests pass. It is only the hg tests that fail.

niclasr:
Here is an output from make integration run with pure 64-bit binaries.
https://gist.github.com/niclasr/2ff795325001b1b2d511b918192d4a28

jcea:
Hitting this too. Any progress?. Thanks.

wez:
@niclasr sorry, this fell through the cracks as the end of the year wound down for me.   That test outpit looks promising!  Would you mind rebasing and updating this pull request?

jcea:
Hitting this too. Any progress?. Thanks.

niclasr:
@wez Not your fault that nothing has happened in a long time. I have updated it now with a fix for
catching the root delete event when "rm -r" is used. The only thing that fails now are the scm tests
and the unicode tests. I don't know how to fix the scm tests and I don't know how important the unicode tests are.

niclasr:
@wez I have now rebased the pull request and removed the conflicting commits. Is it okay merge this now?

wez:
I'll review and (probably!) merge it today, thanks!

wez:
as a heads up, I have a stack of pending changes around w_string/w_string_piece that are likely to land some time this week.  There's a reasonable chance that they may accidentally break this but it should be straightforward to fixup.

watcher/portfs.cpp
```diff
@@ -0,0 +1,323 @@
...
+    const std::shared_ptr<w_root_t>& root,
+    struct watchman_dir* dir,
+    struct timeval,
+    const char* path) {
+  struct stat st;
+
+  auto osdir = w_dir_open(path);
+
+  if (fstat(osdir->getFd(), &st) == -1) {
+    if (w_string_equal(dir->getFullPath(), root->root_path)) {
```
wez: this can/should be:

```
if (dir->getFullPath() == root->root_path) {
```

watcher/portfs.cpp
```diff
@@ -0,0 +1,323 @@
...
+    w_expand_flags(pflags, pe, flags_label, sizeof(flags_label));
+    w_log(
+        W_LOG_DBG,
+        "port: %s [0x%x %s]\n",
+        f->port_file.fo_name,
+        pe,
+        flags_label);
+
+    if ((pe & (FILE_RENAME_FROM | UNMOUNTED | MOUNTEDOVER | FILE_DELETE)) &&
+        w_string_equal(f->name, root->root_path)) {
```
wez: `f->name == root->root_path`

watcher/portfs.cpp
```diff
@@ -0,0 +1,323 @@
...
+          (void*)rawFile)) {
+    watchman::log(
+        watchman::ERR, "port_associate failed in PortFS root delete watch");
+    return false;
+  }
+  return true;
+}
+
+bool PortFSWatcher::startWatchFile(struct watchman_file* file) {
+  auto name = w_dir_path_cat_str(file->parent, file->getName());
```
wez: I broken this out from under you in https://github.com/facebook/watchman/commit/085866203c94f1d1b38eefa6f047b2eb16dbcb0d

I think you want this instead:

```
auto name = file->parent->getFullPathToChild(file->getName());
```

log.cpp
```diff
@@ -33,6 +33,17 @@ void write_stderr(const String& str, Strings&&... strings) {
 }
 }
 
+#if defined(__sun) && !defined(HAVE_DPRINTF)
```
wez: Rather than adding this here (and the check to configure), I think you can replace the `dprintf` call completely.  There are only two calls and they are down around line 238 of this file:

```
   if (si) {
     dprintf(STDERR_FILENO,
         "Terminating due to signal %d %s "	
         "generated by pid=%d uid=%d. %s (%p)\n",
         signo, w_strsignal(signo), si->si_pid, si->si_uid,
         reason, si->si_value.sival_ptr);
   } else {	
     dprintf(STDERR_FILENO, "Terminating due to signal %d %s. %s\n",	  
       signo, w_strsignal(signo), reason);	    
   }	  
```

Can be replaced with something like:

```
if (si) {
  auto msg = watchman::to<std::string>(
        "Terminating due to signal ", signo, " ", w_strsignal(signo),
         " generated by pid=", si->si_pid, " uid=", si->si_uid,  " ",
         reason, " (", si->si_value.sival_ptr, ")\n");
  ignore_result(write(fd, msg.data(), msg.size()));
} else {
auto msg = watchman::to<std::string>(
        "Terminating due to signal ", signo, " ", w_strsignal(signo),
         " ", reason, "\n")
  ignore_result(write(fd, msg.data(), msg.size()));
}
```

niclasr:
changes are now done and they build.

jcea:
Great. What is the timeschedule for a formal release Illumos compatible?. Thanks!.

PR #538, opened 2017-10-04
./configure reports missing openssl headers

fbauzac:
Fail during the configure step when there are missing openssl headers,
with a helpful error message.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,10 @@ AS_IF([test "x$with_folly" != "x"], [
 	)
 ])
 
+dnl Openssl headers are required.
+AC_CHECK_HEADER([openssl/sha.h], [], [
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -45,6 +45,10 @@ AS_IF([test "x$with_folly" != "x"], [
 	)
 ])
 
+dnl Openssl headers are required.
+AC_CHECK_HEADER([openssl/sha.h], [], [
+  AC_MSG_FAILURE([openssl/sha.h is required.  Try e.g. "dnf install openssl-devel"])
```
wez: Can you make these instructions refer to the installation docs instead, and then expand on the various platform specific ways to install the required packages there?

You can find info about updating the docs here: https://facebook.github.io/watchman/contributing.html#dont-forget-the-docs

PR #586, opened 2018-03-12
Update windows download link

AdrianoCahete:


wez:
thanks; landed the equivalent change in:
https://github.com/facebook/watchman/commit/94e66865386e844f2cffe52e355a94c96562d2e3

PR #614, opened 2018-05-31
Fix macOS Python 2.6 builds on Travis CI

strager:
get-pip.py dropped support for Python 2.6. This causes our CI scripts
which set up Python 2.6 on macOS to fail.

pyenv has a workaround: when installing Python 2.6, use a version of
get-pip.py which is compatible with Python 2.6. This workaround requires
a recent version of pyenv. Unfortunately, `brew install pyenv` refuses
to upgrade pyenv.

Install the latest version of pyenv by running `brew upgrade pyenv`
instead of `brew install pyenv`. This causes pyenv to use a working
version of get-pip.py, fixing the setup of Python 2.6.

### DIFF
--- a/travis/deps.sh
+++ b/travis/deps.sh
@@ -1,8 +1,61 @@
 #!/bin/bash
 set -x
+
+brew_install_latest_stable() {
+  local packages=("${@}")
+  set +x
+  local packages_to_install=()
+  local packages_to_upgrade=()
+  for package in "${packages[@]}"; do
+    case "$(brew_package_installed_status "${package}")" in
+      not_installed)
+        packages_to_install+=("${package}")
+        ;;
+      installed_but_outdated)
+        packages_to_upgrade+=("${package}")
+        ;;
+      installed_and_up_to_date)
+        ;;
+      *)
+        printf 'error: unknown package installed status\n' >&2
+        return 1
+        ;;
+    esac
+  done
+  set -x
+  if [ "${#packages_to_upgrade[@]}" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


strager:
Somehow more things are broken. I need to see if my changes are the cause or if there are new issues.

strager:
I ran Travis CI from my fork and the macOS builds pass: https://travis-ci.org/strager/watchman/builds/386420455

How do I retry continuous-integration/travis-ci/pr on this pull request?

wez:
Thanks for this!

I typically iterate by force-pushing to the fork that I used to submit the PR; that should update the PR and kick off the builds again.  If you don't like force pushing, you can add more commits to your branch and push them and they'll show up here.  When we import them via `bunnylol oss watchman` they'll get squashed into a single phabricator diff.

wez:
This is looking good (I think the circle issues are the ones you already fixed in the other PR) but I wonder if we should convert deps.sh into a python script now?  I'm not a huge fan of complicated bash scripts.  Would you mind converting it?  I realize that this might be a bit chicken-and-egg given that that script is trying to figure out which python version to run with :-p

For some more context: In the medium term, one of the things I'd like to see happen is to pull in something like https://github.com/facebookexperimental/eden/blob/master/getdeps.py so that we can make the opensource build pull in `folly` as a dep.   That's a bigger effort and doesn't have to happen as part of this pull request, I'm just sharing that direction and that script in case it is helpful here.

strager:
@wez, my goal with this PR is to fix CI enough so that I can submit real PRs. =] I'm not signing up to maintain the CI scripts.

Issue #369, opened 2016-09-23
Watchman hangs silently when the state directory is corrupted


sandlerr:
Hi maintainers,
Thanks for producing this extremely useful tool! I'm using it for an Ember.JS project with very many files on OSX 10.11.6. Recently my `$USER-state` directory must have gone into a bad state, causing many watchman commands to hang on v4.7.0 and v4.6.0, although `watchman version` worked only on v4.6.0.

I didn't save the corrupt data but if anyone is interested in at least making this fail loudly I have `gdb` output:

```
Thread 1 (Thread 0x1803 of process 69116):
#0  0x00007fff882eb362 in read () from /usr/lib/system/libsystem_kernel.dylib
#1  0x00000001034318df in fill_buffer (jr=0x7fff5c7d01a0, stm=0x7fac7b401f50) at json.c:70
#2  0x0000000103431e63 in read_and_detect_pdu (jr=0x100000, stm=0x7fff5c7d01a0, jerr=0x7fac7b401f50) at json.c:247
#3  0x000000010343194b in w_json_buffer_passthru (jr=0x7fac7b401f50, output_pdu=is_json_pretty, output_pdu_buf=0x7fff5c7d01a0, stm=0x100000) at json.c:406
#4  0x0000000103443f94 in try_command (cmd=<optimized out>, timeout=<optimized out>) at main.c:862
#5  0x0000000103443234 in main (argc=<optimized out>, argv=<optimized out>) at main.c:1043
```

and `strace` output: https://gist.github.com/sandlerr/6074dbee1663524593e553e39088430d

paramaggarwal:
Try this: https://github.com/ember-cli/ember-cli/issues/6275#issuecomment-248399996

sandlerr:
Thanks @paramaggarwal I should have mentioned that this did indeed solve the problem.

sandlerr:
I suppose this is another instance of https://github.com/ember-cli/ember-cli/issues/6259 and https://github.com/facebook/watchman/issues/358 etc.

wez:
Most likely the same issue as #358, which is something we can't easily detect since launchd is the thing that didn't behave as requested.

Closing this in favor of #358

PR #634, opened 2018-08-17
Fixes build on FreeBSD Issue #369

ndowens:


FileDescriptor.cpp
```diff
@@ -178,7 +177,7 @@ static void checkCanonicalBaseName(const char *path) {
         to<std::string>("checkCanonicalBaseName(", path, "): (", name,
                         ") doesn't match canonical base (", base, ")"));
   }
-#else
+else
```
wez: this causes the __APPLE__/macOS build to fail:

https://travis-ci.org/facebook/watchman/jobs/417265100#L537

Was this an intentional change?

It's not super clear to me how this fixes the freebsd build, because the conditional

stream_unix.cpp
```diff
@@ -77,6 +77,7 @@ class UnixStream : public watchman_stream {
   struct ucred cred;
 #elif defined(LOCAL_PEERCRED)
   struct xucred cred;
+  int SOL_LOCAL = 0;
```
wez: It's probably "safer" to make this:

```
#ifndef SOL_LOCAL
#define SOL_LOCAL 0
#endif
```

PR #635, opened 2018-08-17
Tell appveyor to install folly

wez:
We want to add a dep on folly, so the first step is to make sure
that the OSS CI can fetch and build it.  On Windows we can pull
in the head version of folly using vcpkg.  Let's see if that works.


appveyor.yml
```diff
@@ -14,10 +14,13 @@ init:
 
 install:
   - python -V
+  - vcpkg install folly:x64-windows --head
```
ras0219-msft: Note that because you're caching the installed folder, this won't automatically update folly every time. If you want to pull down and build the latest folly on every build, you probably want to add a `vcpkg remove folly:x64-windows` above.

wez: Thanks for the insight!  We seem to have a blocker at the moment: the folly vpkg control file has a hard dep on bzip and that site is no longer available, rendering folly unbuildable via vcpkg.

PR #645, opened 2018-09-08
Return byte[]s for non-UTF-8 file paths.

### LINKED ISSUE #643: Java bindings assume strings are utf-8
Despite the watchman docs noting that in bser strings may not be utf-8, and are unknown encoding due to the nature of file system apis, currently the java deserializer assumes utf-8:

https://github.com/facebook/watchman/blob/master/java/src/com/facebook/watchman/bser/BserDeserializer.java#L236

This causes exceptions when some rogue process (I haven't been able to find out which yet), writes non-utf-8 file names:

```
Caused by: java.nio.charset.MalformedInputException: Input length = 1
        at java.base/java.nio.charset.CoderResult.throwException(CoderResult.java:274)
        at java.base/java.nio.charset.CharsetDecoder.decode(CharsetDecoder.java:813)
        at com.facebook.watchman.bser.BserDeserializer.deserializeString(BserDeserializer.java:236)
        at com.facebook.watchman.bser.BserDeserializer.deserializeRecursiveWithType(BserDeserializer.java:332)
        at com.facebook.watchman.bser.BserDeserializer.deserializeTemplate(BserDeserializer.java:302)
        at com.facebook.watchman.bser.BserDeserializer.deserializeRecursiveWithType(BserDeserializer.java:338)
```

I suppose this is expected, but it blows up the entire `deserialize`, so it's not really easy/possible for me (AFAICT?) to recover from this. I have to stop my process, go find the rogue file, delete it, and then restart the process.

Is there an alternative way for `BserDeserializer` to handle this that would let me continue processing? Maybe just log an error and return `null`? I know that is also surprising, but knowing name is nullable is at least a check I can perform.

stephenh:
The watchman docs note that, due to file system APIs being
encoding-less, we can't assume the bser strings are UTF-8.

This changes the bser string deserializer to fallback to returning
a byte[] when UTF-8 deserialization fails.

This lets client to a cast/instanceof check on String or byte[]
and decide how they want to handle the value.

Fixes #643.


mkillianey:
@stephenh  - How is this working out for you in practice?  Once you have the byte array, are you finding a clear way to do something useful with it from Java, or was that a dead-end?  Do you still think it's worth taking this change, or does it cause other headaches elsewhere?

stephenh:
@mkillianey seems good so far, but I am technically checking for "if byte[]" and then skipping those files, and not doing anything with them. This is mostly out of laziness/lack of time to update my app's wire protocol + file system callpaths/etc. to handle "oh right this is a string or a byte[]".

So, no, I don't have a concrete data point that I followed pulled the byte[] thread all the way back into file system APIs and had it work out. (Also the Java file sytems APIs I'm using all take strings, of course, so I'd have to use jnr-poxis equivalents, which I currently do for a few things/special cases like symlinks, but generally try to avoid.)

That said, I did have a user with non-utf-8 paths (based in the UK) that would have triggered my app's previous "blow up w/exception" behavior, so while silently skipping the paths is not necessarily the best right now, it's at least better than failing the entire program.

PR #666, opened 2019-01-09
Add more help to libtoolize fail in autogen.sh to make consistent with pkg-config help

yanneves:
Ran into this tonight, figure this minor change could save some other users two minutes when building from source on a new system.

On missing `libtool` dependency, expanded output from `autogen.sh` from..

```bash
your system lacks libtoolize
```

to..

```bash
your system lacks libtoolize
please install the libtool package for your system.
```

This is borrowed from the same format used for missing `pkg-config` dependency, where output from `autogen.sh` helpfully states..

```bash
pkg-config appears to be missing (not available to autoconf tools)
please install the pkg-config package for your system.
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -8,7 +8,7 @@ if test -d config.status ; then
 fi
 
 for p in libtoolize glibtoolize FAIL; do
-  test "$p" = FAIL && { echo "your system lacks libtoolize" 1>&2; exit 1; } || :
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this, however, we're in the middle of replacing autoconf with cmake; I have some diffs pending some tests that will make their way over to the github repo in the next couple of weeks and then the autoconf build stuff will be deleted, and the libtool dependency will be removed along with it.

yanneves:
Thanks @wez, in that case I'll close this pull request. Good luck with the updates.

PR #670, opened 2019-01-20
Check CreateFileW() return value correctly

PawanKartikS:
CreateFileW() returns "INVALID_HANDLE_VALUE" if the handle is invalid.  Using the `if(!dir_handle)` does not correctly check for invalid handle value.

Code that reproduces this behaviour:
````
HANDLE x = INVALID_HANDLE_VALUE;
if (x == INVALID_HANDLE_VALUE)
{
	std::cerr << "Comparison with INVALID_HANDLE_VALUE." << std::endl;
}

if (!x)
{
	std::cerr << "Using the '!' operator." << std::endl;
}

````

Output for the above code: `Comparison with INVALID_HANDLE_VALUE`.

### DIFF
--- a/watcher/win32.cpp
+++ b/watcher/win32.cpp
@@ -68,7 +68,7 @@ WinWatcher::WinWatcher(w_root_t* root)
       OPEN_EXISTING,
       FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OVERLAPPED,
       nullptr);
-  if (!dir_handle) {
+  if (dir_handle == INVALID_HANDLE_VALUE) {
     throw std::runtime_error(
         std::string("failed to open dir ") + root->root_path.c_str() + ": " +
         win32_strerror(GetLastError()));
--- a/watcher/win32.cpp
+++ b/watcher/win32.cpp
@@ -32,7 +32,7 @@ struct Item {
 
 struct WinWatcher : public Watcher {
   HANDLE ping{INVALID_HANDLE_VALUE}, olapEvent{INVALID_HANDLE_VALUE};
-  HANDLE dir_handle{INVALID_HANDLE_VALUE};
+  FileDescriptor dir_handle(INVALID_HANDLE_VALUE);
 
   std::condition_variable cond;
   watchman::Synchronized<std::list<Item>, std::mutex> changedItems;
@@ -63,7 +63,7 @@ WinWatcher::WinWatcher(w_root_t* root)
 
   // Create an overlapped handle so that we can avoid blocking forever
   // in ReadDirectoryChangesW
-  FileDescriptor dir_handle(intptr_t(CreateFileW(
+  dir_handle = intptr_t(CreateFileW(
     wpath.c_str(),
     GENERIC_READ,
     FILE_SHARE_READ | FILE_SHARE_DELETE | FILE_SHARE_WRITE,
@@ -71,7 +71,7 @@ WinWatcher::WinWatcher(w_root_t* root)
     OPEN_EXISTING,
     FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OVERLAPPED,
     nullptr
-  )));
+  ));
 
   if (!dir_handle) {
     throw std::runtime_error(
--- a/watcher/win32.cpp
+++ b/watcher/win32.cpp
@@ -70,7 +70,7 @@ WinWatcher::WinWatcher(w_root_t* root)
     nullptr
   )));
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PawanKartikS** [comment]
Made the changes you've asked. Sorry, took me a while because I am missing out a bunch of dependencies and CMake was refusing to build.

**Comment by wez** [comment]
I did just change things dramatically in master for the build; is something like this should fetch all the required deps, assuming that you have vcpkg installed in `c:/tools/vcpkg`:

```
python getdeps.py --install-deps
cmake -G "Visual Studio 15 2017 Win64" -DCMAKE_TOOLCHAIN_FILE=c:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake .
```

**Comment by PawanKartikS** [comment]
Btw, is it `using namespace watchman;` or `using watchman::FileDescriptor;`? I'm having intellisense issues with the former. Also, `stream_win.cpp` uses  `using watchman::FileDescriptor;`
https://github.com/facebook/watchman/blob/master/stream_win.cpp#L7

**Comment by PawanKartikS** [comment]
Also, `vcpkg` is having issues detecting and using VS 2019 (tried building from source by cloning vcpkg:master). This is preventing my CMake from building the watchman from source.

vcpkg error:
```
Could not find Visual Studio. VS2015 or VS2017 (with C++) needs to be installed.
At C:\Users\Satya\Documents\Repos\vcpkg\scripts\bootstrap.ps1:143 char:9
+         throw "Could not find Visual Studio. VS2015 or VS2017 (with C ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Could not find ...o be installed.:String) [], RuntimeException
    + FullyQualifiedErrorId : Could not find Visual Studio. VS2015 or VS2017 (with C++) needs to be installed.
```

Apparently `vcpkg` uses `vswhere` to detect active/installed VS copy. From the build script, looks like `vswhere` is failing.

**Comment by PawanKartikS** [comment]
Btw, there's another issue. In the python file, `isdir` will fail for files. It returns True only for dirs, and false for files.
It must be replaced by `os.path.exists()`.

**Comment by PawanKartikS** [comment]
Btw, os.path.isdir() returns False for files, True for dirs.

```py
>>> os.path.isdir("C:/tools/vcpkg.exe")
False
>>> os.path.exists("C:/tools/vcpkg.exe")
True
```

wez:
> Btw, is it `using namespace watchman;` or `using watchman::FileDescriptor;`? I'm having intellisense issues with the former. Also, `stream_win.cpp` uses `using watchman::FileDescriptor;`
> https://github.com/facebook/watchman/blob/master/stream_win.cpp#L7

You can use either; the latter is more explicit and the former is more broad; use whichever one you want!

wez:
> Btw, there's another issue. In the python file, `isdir` will fail for files. It returns True only for dirs, and false for files.
> It must be replaced by `os.path.exists()`.

Which python file are you seeing that in, `getdeps.py`?  You could open a second pull request to correct that issue and we can dig into it there :)

PawanKartikS:
> > Btw, is it `using namespace watchman;` or `using watchman::FileDescriptor;`? I'm having intellisense issues with the former. Also, `stream_win.cpp` uses `using watchman::FileDescriptor;`
> > https://github.com/facebook/watchman/blob/master/stream_win.cpp#L7
> 
> You can use either; the latter is more explicit and the former is more broad; use whichever one you want!

Well, using either of them broke the build. I'm right now making changes to the `getdeps.py` so that I get all the dependencies installed and will then open a second PR for it. I'd like to resolve these issues before I get to the `FileDescriptor` so that I can run a full build on my machine.

PawanKartikS:
@wez Are you sure `dir_handle.handle()` is right? I get this error everywhere I invoke `.handle()`. MSDN says - `intptr_t - Stores a pointer (or HANDLE) on both Win32 and Win64 platforms.` `FileDescriptor.cpp` seems to be using the cast `(HANDLE)` in few places.

So should I cast it?

PS: Sorry for the inconvenience! `getdeps.py` now seems to detect `vcpkg` after the tweak, but quits midway during the dependency installation. I'm blind here and am unable to run a full build on my machine.

untitled

PawanKartikS:
Dependencies are not being built and set up correctly by VS19/vcpkg. Microsoft has reached out to me and I'm in touch with the VS team to further investigate and resolve this. 

Should this PR be closed or let it remain as it is?

PR #694, opened 2019-04-01
Allow configuring WATCHMAN_STATE_DIR at build time

strager:
Watchman's old autoconf-based build system allowed customizing the value of WATCHMAN_STATE_DIR, or disabling WATCHMAN_STATE_DIR entirely. Watchman's new CMake-based build system does not. Allow users to configure WATCHMAN_STATE_DIR when using the CMake build system.


genevievehelsel:
This PR is a bit stale - is this still something that you'd like us to review? Could you either please close out the PR if its no longer needed or rebase the change if its something that still should be reviewed :)

PR #695, opened 2019-04-01
Allow building without Python

strager:
Watchman's CMake build system fails if a Python interpreter is not found. This is unnecessary; Watchman's build does not require Python. Make Python an optional dependency instead of a required dependency.


wez:
Implemented in https://github.com/facebook/watchman/commit/e9ab4acba23109add581494055248b519523f174

PR #706, opened 2019-05-21
WIP: Adopt new getdeps for posix oss builds

wez:
Try cutting the CI over to the new build runner.
I'm deliberately avoiding appveyor for the moment because we
need to wrangle some dynamic deps as part of its artifact build
and we're not ready for that.  It's better to leave it broken
for the time being so that the download link continues to point
to the last successful build.


wez:
succeeded by https://github.com/facebook/watchman/pull/743

PR #712, opened 2019-06-19
[node] Fix deprecated new Buffer call

kassens:
See note at:
https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe

`new Buffer(size)` is deprecated and should be replaced by the explicit
`Buffer.alloc(size)` or `Buffer.allocUnsafe(size)` which doesn't zero
out the memory.


strager:
Looks good to me!

PR #727, opened 2019-06-26
Add support for KDE neon

tiguchi:
KDE neon is based on the latest Ubuntu LTS version (currently 18.04), therefore Debian and Ubuntu dependency installation works here, too.

`./autogen.sh`, `make` and `make install` work fine on KDE neon with this modification. Tested on "KDE neon User Edition 5.16"


wez:
I'm replacing autogen.sh and getdeps.py with something else shortly

PR #738, opened 2019-08-14
Included `pkg-config` package

arthurRuf:
Included `pkg-config` package required to Installing from source section


wez:
we'll be restructuring this section of the docs soon

Issue #710, opened 2019-06-13
Null dereference


QiAnXinCodeSafe:
Null pointer value stored to 'o'
https://github.com/facebook/watchman/blob/c8a74cdce0ad54763a6104ec84aa06bfcf7473f2/opt.cpp#L188

Assuming the condition is false
https://github.com/facebook/watchman/blob/c8a74cdce0ad54763a6104ec84aa06bfcf7473f2/opt.cpp#L190

Access to field 'is_daemon' results in a dereference of a null pointer (loaded from variable 'o')
https://github.com/facebook/watchman/blob/c8a74cdce0ad54763a6104ec84aa06bfcf7473f2/opt.cpp#L197

wez:
This condition cannot be met; `getopt_long` would have returned `?` and we would have already returned earlier.

PR #742, opened 2019-09-10
Check if dereferencing a nullptr.

PawanKartikS:
Pointer to `watchman_getopt` can be null if certain conditions fail. This commit implements the required checks.


wez:
Is this PR in response to https://github.com/facebook/watchman/issues/710 ?
If so, I don't believe that it can be hit in practice and we don't need to change this code.
However, if I have overlooked something, please show how to reach that section of code with `o == nullptr`!

PR #748, opened 2019-10-01
ci/circle: use up-to-date cmake from apt

Artoria2e5:
The ci is failing to build because the self-built "newer cmake" is not new enough. In fact, the cmake in the ubuntu 18.04 apt repository is at 3.10.2. Let's use that one instead.

Bumped debian up to buster because that's the newer stable Debian release with cmake 3.13.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,25 +5,12 @@ defaults: &defaults
         name: Install dependencies
         command: |
           apt-get update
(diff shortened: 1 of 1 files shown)


Artoria2e5:
fucking hell lads how many CIs are you using

wez:
Thanks for looking at this!  However, we're standardizing on Github Actions and will be removing the other CI configs (the others are too flakey, too slow and have build timeouts that are too low to build our full feature set).

We need to revise instructions on how to manually build things, but the gist of it is that you can run:

```bash
./build/fbcode_builder/getdeps.py build watchman
mkdir /tmp/install-root
./build/fbcode_builder/getdeps.py fixup-dyn-deps watchman /tmp/install-root
```

to build and deploy the executables under `/tmp/install-root`

PR #749, opened 2019-10-01
List all the boost components used to silence CMake errors

Artoria2e5:
TravisCI cmake is failing due to these things. Shut it up.


wez:
Thanks for looking at this!  Travis is failing for a whole bunch of reasons; we've migrated to GH actions instead and will be deleting the travis config shortly

PR #754, opened 2019-10-12
Fixing Travis build, remove OS X, up to 18.04

### LINKED ISSUE #752: Travis build failures
Travis builds for `master` are red for a long time https://travis-ci.org/facebook/watchman/builds Does that mean that `watchman` builds are broken on Linux?

**EDIT**: The build was broken in [Feb 2019](https://github.com/facebook/watchman/commits/master?after=658e9ec680fe7691e1fe1d136180b282511a301a+925) - most likely by https://github.com/facebook/watchman/commit/631bdda241a136d975883de7d78aacce59dc0721.

abitrolly:
Just a test to see if it could fix #752

Ref. https://docs.travis-ci.com/user/reference/bionic/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,80 +1,19 @@
 language: c
 
-# use container infra for linux
-dist: xenial
-sudo: false
-
-addons:
-  apt:
-    sources:
-      - deadsnakes
-      - ubuntu-toolchain-r-test
-    packages:
-      - cmake
-      - python2.7
-      - python2.7-dev
-      - python3.5
-      - python3.5-dev
-      - python3.6
-      - python3.6-dev
-      - gcc-5
-      - g++-5
-      - gcc-6
-      - g++-6
-      - autoconf-archive
-      - bison
-      - build-essential
-      - ccache
-      - cmake
-      - curl
-      - flex
-      - git
-      - gperf
-      - joe
-      - libboost-all-dev
-      - libcap-dev
-      - libdouble-conversion-dev
-      - libevent-dev
-      - libgflags-dev
-      - libgoogle-glog-dev
-      - libpthread-stubs0-dev
-      - libkrb5-dev
-      - libpcre3-dev
-      - libnuma-dev
-      - libsasl2-dev
-      - libsnappy-dev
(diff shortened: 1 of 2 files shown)


wez:
Thanks for looking at this! However, we're standardizing on Github Actions and will be removing the other CI configs (the others are too flakey, too slow and have build timeouts that are too low to build our full feature set).

We need to revise instructions on how to manually build things, but the gist of it is that you can run:

```
./build/fbcode_builder/getdeps.py build watchman
mkdir /tmp/install-root
./build/fbcode_builder/getdeps.py fixup-dyn-deps watchman /tmp/install-root
```

to build and deploy the executables under `/tmp/install-root`

abitrolly:
Indeed. Travis could not complete because of timeout.
Is it possible to run GitHub Actions locally? 

I was using instruction from https://facebook.github.io/watchman/docs/install.html#installing-from-source - there is no other way to install on Linus except installing all dependencies and compiling everything. Much better UX is to provide Linux binaries for download.

wez:
> I was using instruction from https://facebook.github.io/watchman/docs/install.html#installing-from-source - there is no other way to install on Linus except installing all dependencies and compiling everything. Much better UX is to provide Linux binaries for download.

With GH actions we do now have the ability to offer linux binaries.  It's a bit clunky, but here's how to get to them:

1. View list of passing builds at
https://github.com/facebook/watchman/actions?query=is%3Acompleted+event%3Apush+is%3Asuccess
2. Click on the most recent one
3. In the top right corner there is an "Artifacts" drop down; click it and it will show you a "watchman" download (see https://github.com/actions/upload-artifact/issues/2)
4. Inside that zip file you'll find binaries for all three platforms

I'd like to see this streamlined in the future but it depends on https://github.com/actions/upload-artifact/issues/21

abitrolly:
172Mb for Linux binary. That's huge to be run as a resident process compared, for example, to `watchexec` which is only 3Mb. 

I see there are now updated instructions. https://facebook.github.io/watchman/docs/install.html#binary-downloads-for-linux-macos-and-windows-beta That's nice.

PR #773, opened 2020-01-09
Upload the artifacts to public Github Releases

### LINKED ISSUE #765: Installation - Github Actions is only accessable if logged in on github.com
When on the installation page https://facebook.github.io/watchman/docs/install.html and following the install instructions below

image
I come to a 404 page on github

image

Clearly, this is an error on githubs side, they should not give me a 404 if I'm not logged in, they should ask me to log in.
But until they have sorted that out, it would be nice if watchman supplied a workaround.

**Suggestion** Clarify on the install page https://facebook.github.io/watchman/docs/install.html that you need to be logged in to github to be able to access the actions section.

kumarharsh:
This step uses the [Github Releases](https://github.com/marketplace/actions/github-releases) action.

**PLEASE NOTE**: While merging this PR, make sure to also add a `GITHUB_TOKEN` environment variable for this repository in the repository Settings page.

Closes #765

### DIFF
--- a/build/fbcode_builder/getdeps.py
+++ b/build/fbcode_builder/getdeps.py
@@ -663,6 +663,13 @@ jobs:
         out.write("        name: %s\n" % manifest.name)
         out.write("        path: _artifacts\n")
 
+        out.write("    - name: Copy release\n")
+        out.write("    - uses: fnkr/github-action-ghr@v1\n")
+        out.write("      env:\n")
+        out.write("        GHR_PATH: _artifacts\n")
+        out.write("        GHR_COMPRESS: zip\n")
+        out.write("        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}\n")
+        
         out.write("    - name: Test %s\n" % manifest.name)
         out.write(
             "      run: python build/fbcode_builder/getdeps.py test --src-dir=. %s\n"


wez:
Thanks for looking at this!  There are a couple of complexities here:

* Changes to fbcode_builder will get mirrored back into at least a dozen of our various opensource projects, so we need to consider the impact on those; it's probably fine, but plays into the next point
* I don't think that we would trust anything other than the official GH provided actions with our access token
* Most of the releases related actions I've looked at will make a new release each time they trigger.  We push several commits per day and this will get noisy very fast.  In addition, we don't manually tag releases across most of the projects because the source of truth is a rolling release from a monorepo that is then mirrored out to GitHub.  There isn't a concept of a release that it would make sense to map to this; that's one of the reasons that we haven't made a watchman release in a while.

The "solution" I came up for this in a personal project was to create a dummy "nightly" release and then have the CI upload the artifacts to that each they run, replacing the prior version.  It's a bit of a hack really, and is a bit awkward to apply to all our projects in a scalable way.

I'll raise this with some folks internally to see if what we can come up with to solve this holistically!

kumarharsh:
Thanks for your detailed response @wez. While creating the PR, I had guessed some reasons about why this wasn't already done, and seems like I was pretty much bang on 😄

I agree about not exposing the GH token to third-party actions, and it's weird that Github themselves haven't provided a solution for this.

I'd have liked if there were atleast some tagged releases done for watchman, but as it's a rolling release, I understand that this would be kind of hard. Although it's still a question as to why the Watchman distributed via Brew is on 4.9.0, when the one gotten from actions is on 4.9.4. Maybe the release process is not performing well?

Ultimately, I just wanted to get the ball rolling on improving the release process as there are a bunch of issues around it, so it's heartening to see your response 🙂

wez:
Obsoleted by https://github.com/facebook/watchman/pull/828

PR #774, opened 2020-01-09
docs: s/OS X/macOS

brandonkal:
This is a documentation change only to use the preferred name for the macOS operating system.

Justification:
It has been four years since Apple renamed **OS X** to **macOS**.
Also note that the bottle Homebrew now only supports [macOS Sierra](https://github.com/Homebrew/homebrew-core/blob/ff9bd877e442f32790267f114234413b9ee540a5/Formula/watchman.rb#L19-L22) and later (which is when Apple made the change to the new name). A Linux Homebrew formula is [also available from core](https://github.com/Homebrew/linuxbrew-core/blob/master/Formula/watchman.rb#L22-L23) so I made that change as well.

### DIFF
--- a/opendir.cpp
+++ b/opendir.cpp
@@ -76,7 +76,7 @@ class DirHandle : public watchman_dir_handle {
 static DIR* opendir_nofollow(const char* path) {
   auto fd = openFileHandle(path, OpenFileHandleOptions::strictOpenDir());
 #if !defined(HAVE_FDOPENDIR) || defined(__APPLE__)
-  /* fdopendir doesn't work on earlier versions OS X, and we don't
+  /* fdopendir doesn't work on earlier versions macOS, and we don't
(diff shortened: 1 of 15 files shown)


wez:
Thanks for taking this on!
regarding linux brew: I haven't personally tried to use it to install watchman; does that work?  I'd rather not push folks towards without having at least someone chime in with positive thoughts about it :-)

brandonkal:
Yes, I installed via brew on Linux and the binary is working correctly. @wez

brandonkal:
Neat. It's unfortunate the merge bot doesn't include a Co-authored-by attribution.

wez:
@brandonkal you show as the author in the commit that got squashed and pushed.  Is there some metadata that got mangled?

brandonkal:
@wez I see it now. Everything looks good.

PR #788, opened 2020-03-19
Migration of docs to Docusaurus

fanny:
## Motivation

Hello there, as discussed in #787  we would like to dogfood docusaurus v2 for watchman. The development experience for v2 is much better because client side rendering is also enabled. The website is also seems to be very performant (its a single page app now but also staticly rendered for SEO)

## Demo

0F6EF9E0-BFB8-4065-A7B4-420E53E28C59_1_105_c

## How to Test
`cd website && npm install && npm start`

cc @yangshun


website/src/pages/index.js
```diff
@@ -0,0 +1,78 @@
...
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import styles from './styles.module.css';
+
+function Home() {
+  const context = useDocusaurusContext();
+  const {siteConfig = {}} = context;
+  return (
+    <Layout
+      title={`Hello from ${siteConfig.title}`}
```
wez: should this match the `<title>` from the current site?  If so, that is set to `Watchman A file watching service | Watchman` right now

website/src/pages/index.js
```diff
@@ -0,0 +1,78 @@
...
+      description="Description will go into a meta tag in <head />">
```
wez: similarly for this?

website/docs/trigger.md
```diff
@@ -209,7 +206,7 @@ settings:
 * `name` is set to the `triggername`
 * `command` is set to the `<cmd>` list
 * `expression` is generated from the `<patterns>` list using the rules laid
-  out in [Simple Pattern Syntax](/watchman/docs/simple-query.html)
+  out in [Simple Pattern Syntax](/docs/simple-query)
```
wez: A lot of these links have been changed from `/watchman/docs/` to just `/docs/`.  I think that makes sense for local testing, but when these are deployed to github that isn't right (eg: https://facebook.github.io/watchman/docs/install.html is where the install docs get served up).

Can these be changed to omit the paths?  For example could this one could perhaps just be `simple-query` instead of `/docs/simple-query`?

website/docusaurus.config.js
```diff
@@ -0,0 +1,65 @@
...
+    },
+  },
+  presets: [
+    [
+      '@docusaurus/preset-classic',
+      {
+        docs: {
+          sidebarPath: require.resolve('./sidebars.js'),
+          editUrl:
+            'https://github.com/facebook/docusaurus/edit/master/website/',
```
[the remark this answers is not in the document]
fanny: Yes, sorry

fanny:
> This looks great, thanks!
> 
> One thing that worth calling out; take a look at the install docs; there's a portion of the instructions that looks like this:
> 
> ```
> $ git clone https://github.com/facebook/watchman.git -b {{ site.data.current_release.tag }} --depth 1
> ```
> 
> in the current jekyll based docs that text renders like this here:
> 
> https://facebook.github.io/watchman/docs/install.html#installing-from-source
> 
> ```
> $ git clone https://github.com/facebook/watchman.git -b v4.9.0 --depth 1
> ```
> 
> is something like this possible with docusaurus? If not, we'll need to replace this with the actual version number. The original intent of this was to factor out the current version number to make it easier to update throughout the site, but since we only use it in a single place right now its not that important to preserve this functionality.

Is possible, but you need to develop a versionated docs, that means every release you'll have a folder with all markdown files and pages, like:

```
website
├── sidebars.json        # sidebar for master (next) version
├── docs                 # docs directory for master (next) version
│   ├── foo
│   │   └── bar.md       # https://mysite.com/docs/next/foo/bar
│   └── hello.md         # https://mysite.com/docs/next/hello
├── versions.json        # file to indicate what versions are available
├── versioned_docs
│   ├── version-1.1.0
│   │   ├── foo
│   │   │   └── bar.md   # https://mysite.com/docs/foo/bar
│   │   └── hello.md
│   └── version-1.0.0
│       ├── foo
│       │   └── bar.md   # https://mysite.com/docs/1.0.0/foo/bar
│       └── hello.md
├── versioned_sidebars
│   ├── version-1.1.0-sidebars.json
│   └── version-1.0.0-sidebars.json
├── docusaurus.config.js
└── package.json
```

I don't know if this is your goal.

website/docs/trigger.md
fanny: I tested, isn't works, because the `/`, serves to identify that after the base url, in this case `facebook.github.io`, the route will be `docs/simple-query` if we change this, the docusaurus will cannot understand the path

fanny: Yeah, sorry again I didn't tested using this way. I'll change ASAP.

wez: Thanks; I think the leading `/` might not be good because the docs run out from `/watchman/` and `/` will be the wrong directory.  Is there a way to test with a subdir locally?

wez:
> Is possible, but you need to develop a versionated docs, that means every release you'll have a folder with all markdown files and pages, like:
> 
> ```
> website
> │   └── version-1.0.0

> ```
> 
> I don't know if this is your goal.

Oh, nothing that complicated; this was more like a macro.  We can just drop this and replace it with a manual version number.

website/docs/trigger.md
fanny: Yes, I do thinks. Good observation.

website/docusaurus.config.js
```diff
@@ -0,0 +1,65 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+module.exports = {
+  title: 'Watchman',
+  tagline: 'A file watching service',
+  url: 'https://facebook.github.io/watchman/',
+  baseUrl: '/',
```
wez: I needed to change this to `/watchman/` for the docs to render right; I'll fix this when I import to phabricato!

fanny: Oh Thanks @wez , if you want that i change, let me know. I thought that was not necessary put `/watchman`,  is present in the url.

wez: It's ok; I've amended it into the phabricator diff.  Without this all the styling and js links are relative to `/` and the page looks unhappy.

PR #806, opened 2020-04-26
Update API docs link url in npm package

conartist6:


wez:
This got fixed more globally in the docs; thanks for helping out!

PR #846, opened 2020-08-13
task/add command to cd into extracted release

mudathirlawal:
Hi there,

A user raised this suggestion and I think it is a useful improvement and, therefore, decided to add it. In the "Installation" section of the contributing guide, it is not stated that the last four commands are to be executed in the directory of the extracted release with the name format: **watchman-vYYYY.MM.DD.00-linux**. Meanwhile, including will prevent users from getting unnecessarily stuck, especially users with little or no familiarity with the command line.

My changes include the command to change directory into the extracted release folder just before the commands that need to be run within the directory.

Warm regards.
**_Mudathir Lawal_**


fanzeyi:
Sorry, this is already done in #845. Thanks for contributing.

PR #852, opened 2020-08-31
Add Windows Chocolatey Install Documentation

dgalbraith:
It is now possible to install Watchman on Windows with the
[Watchman package](https://chocolatey.org/packages/watchman).

Updated installation documentation to reflect this and included details
on how to install Watchman on Windows using Chocolatey.


wez (changes requested):
Thanks for contributing!

We're a little cautious about the implications of linking to something that is outside of our control, so I'd suggest some wording changes here:

---

> Watchman is available to install via the [Chocolatey](https://www.chocolatey.org) Windows package manager.  Installation is as simple as:

```powershell
PS C:\> choco install watchman
```

> The package is maintained by the community rather than by Facebook, Inc., so if you experience
issues with installation/uninstallation, you should [contact the package maintainers](https://chocolatey.org/packages/watchman/ContactOwners) for assistance.

---

I don't know how feasible this is, but would it be possible to automatically update the choco package information when we push a release?  It seems desirable to have our GH Actions automation take care of this.  I'm not familiar with the process of updating choco packages; if it requires a manual review then a weekly PR could get annoying for the reviewers.

dgalbraith:
> We're a little cautious about the implications of linking to something that is outside of our control, so I'd suggest some wording changes here:
> 
> > Watchman is available to install via the [Chocolatey](https://www.chocolatey.org) Windows package manager.  Installation is as simple as:
> 
> ```powershell
> PS C:\> choco install watchman
> ```
> 
> > The package is maintained by the community rather than by Facebook, Inc., so if you experience
> > issues with installation/uninstallation, you should [contact the package maintainers](https://chocolatey.org/packages/watchman/ContactOwners) for assistance.

Understood.  Documentation has been updated inline with the suggestions.
 
> I don't know how feasible this is, but would it be possible to automatically update the choco package information when we push a release? It seems desirable to have our GH Actions automation take care of this. I'm not familiar with the process of updating choco packages; if it requires a manual review then a weekly PR could get annoying for the reviewers.

This is already the case.  Whenever a new Windows release is made available on the [releases page](https://github.com/facebook/watchman/releases) the package will be updated and pushed automatically.  This is done through an external CI pipeline that runs once every 24 hours at around 0400 UTC - if a new Windows artifact has been deployed the package will be built and pushed to the Chocolatey community repository.

Should you wish to push updates directly from the GH Actions as part of the build pipeline I would be happy to work with you to get this in place as opposed to using the external pipeline.

wez (approved):

wez:
Thanks for this!

One of the things that I'd like to make progress on is including the python executables in the generated artifacts (need to debug https://github.com/facebook/watchman/pull/856 first) and that might impact the choco packaging.
Longer term, I'd like for us to generate an installer package too.

I think it would be great to automate the choco release updates to be workflows in this repo too, but I don't have a good feel for how much effort that requires right now.

PR #925, opened 2021-07-02
Quick fixes for gcc-9.3.0

MathieuAvila:
This is a quick fix that shows where the issues happens on gcc 3.9.0, and how to work-around them quickly.
Not the finest way to fix, however.


fanzeyi:
Do you mind sharing the errors you run into with gcc 9.3.0? It is not that obvious with the changes in this PR.

MathieuAvila:
superseded by a recent commit

PR #926, opened 2021-07-03
Fix the Rust client library on Windows

leops:
The Rust library doesn't build on Windows since recent versions of Tokio removed the `PollEvented` abstraction. However a new `NamedPipeClient` type has since been introduced that directly implements `AsyncRead` and `AsyncWrite` for a named pipe client, so I modified the `NamedPipe` struct to be a simple convenience wrapper around the tokio implementation.


fanzeyi:
Ohh! Thanks for working on this.

fanzeyi (approved):

PR #938, opened 2021-08-16
Improvements to the website

jangdan:
This PR introduces some minor improvements to the Watchman website. It
1. Removes a (presumably useless) empty `<div>` that I stumbled upon, and it
2. Fixes a problem with the layout for any doc page (["doc_page.html"](https://github.com/facebook/watchman/blob/78fadf453aee21b49c5a99db1ecad1b61434264d/website/_layouts/doc_page.html)) displayed on Google Chrome running on macOS (full browser details: https://www.whatsmybrowser.org/b/PLIF0; It seems setting only `max_width` is not enough).
Here is how it looks: 127 0 0 1_4000_watchman_docs_install html

### DIFF
--- a/website/_includes/nav_docs.html
+++ b/website/_includes/nav_docs.html
@@ -1,5 +1,3 @@
-<div>
-</div>
 <nav class='toc' id="doc_nav"></nav>
 <script>
   var docnavData = [
--- a/website/_sass/_base.scss
+++ b/website/_sass/_base.scss
@@ -918,6 +918,7 @@ a.blockButton {
   
   .wrapper {
     max-width: 1100px;
+    width: 1100px;
   }
     
   .homeWrapper {
@@ -944,6 +945,7 @@ a.blockButton {
   
   .wrapper {
     max-width: 1400px;
+    width: 1400px;
   }
     
   .homeWrapper {


chadaustin:
I can't reproduce this behavior in Chrome on macOS. How do you get it to render like in your screenshot? Is that just how it shows by default, or did you resize the window? Does resizing the window change anything?

jangdan:
I did some "cross-browser window-resizing", and it seems that this error reveals itself when the width of the browser window (`window.innerWidth`) becomes exactly 1200 px. This happens on Firefox as well; It was not a Chrome-specific issue after all.

To add: I think what happened was that my Chrome window just happened to be 1200 px wide when I decided to look at the Watchman docs. This is probably the width you get when you first install Chrome. (I tend to not resize my windows from their defaults.)

chadaustin:
Thanks! I confirmed this works. I'll merge it.

chadaustin:
Sorry, I did the wrong thing with our internal tooling. Let me back this out and import it the other way.

chadaustin:
@jangdan I know this is a small thing, and I could just copy the changes and submit them myself, but can you open a new PR with these changes and then I can import it the proper way? Thanks.

jangdan:
Sure!

jangdan:
@chadaustin #940

PR #942, opened 2021-08-31
Use python-setuptools for Windows builds in CI

mathewhodson:
This resolves the following error when building 'pywatchman.bser' extension:

> error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/


fanzeyi:
Hi!

Thanks for looking into this. I spent some time a few weeks ago trying to fix this as well but I wasn't able to find the root case. Here are my findings in case its useful:

* If I directly build the Python package with `python setup.py` on GitHub Windows environment, it works. ([run](https://github.com/fanzeyi/watchman/actions/runs/1106478369), [commit](https://github.com/fanzeyi/watchman/commit/f39d6628a7f596d3e6ea1f871ed21a973426aa87)).
* If I remove all the CMake stuff and most of the dependencies, it works. ([run](https://github.com/fanzeyi/watchman/actions/runs/1106921379), [commit](https://github.com/fanzeyi/watchman/commit/112e3c811748d24ff3abb41c6181d9a06659d145))
* If I only remove the CMake changes but still let getdeps build the dependencies, it fails. ([run](https://github.com/fanzeyi/watchman/actions/runs/1106846213), [commit](https://github.com/fanzeyi/watchman/commit/f41286a7f65dd8ed63d17d1a543d5075e53ea8b4))

You can find all my attempts at https://github.com/fanzeyi/watchman/actions?query=branch%3Afix-windows-ci++, I do amend and force push all the time so you will have to click on the commit id in the run to see what changes I made.

I suspect somehow having the dependencies messed up setuptools's logic to locate MSVC but I wasn't sure why. Hope this will be helpful. Let me know if you have any questions.

PR #943, opened 2021-09-01
Changes for building in Debian

aweeraman:
Some changes that had to be done in Debian for the build to work:

* Installation of system deps failed as it was not detecting the OS correctly
* Updated the CMake manifest to uncomment the deb dependency as the CMake versions are newer
* fb303 was not getting cloned as it was looking for a non-existent master branch


fanzeyi (approved):
Thanks. This all looks good to me.

Issue #990, opened 2022-01-11
Cannot compile on Linux Mint 20.3 "Una"(Ubuntu 20.04)


franklin-vidal:
Hello,

After trying to compile i get the error:

```
+ PREFIX=/usr/local
+ python3 build/fbcode_builder/getdeps.py build --src-dir=. watchman --project-install-prefix=watchman:/usr/local
Building on {distro=mint, distro_vers=20.3, fb=off, os=linux, test=on}
Assessing boost...
Assessing ninja...
Assessing openssl...
Traceback (most recent call last):
  File "build/fbcode_builder/getdeps.py", line 1130, in <module>
    sys.exit(main())
  File "build/fbcode_builder/getdeps.py", line 1113, in main
    return args.func(args)
  File "build/fbcode_builder/getdeps.py", line 108, in run
    self.run_project_cmd(args, loader, manifest)
  File "build/fbcode_builder/getdeps.py", line 563, in run_project_cmd
    builder = m.create_builder(
  File "/home/frk/Downloads/watchman-2022.01.10.00/build/fbcode_builder/getdeps/manifest.py", line 448, in create_builder
    raise Exception("project %s has no builder for %r" % (self.name, ctx))
Exception: project openssl has no builder for <getdeps.manifest.ManifestContext object at 0x7efefb7110d0>
```

chadaustin:
Fixed by https://github.com/facebook/watchman/commit/7c31dfdb1c8c44d261447e9b03b79492c025a9e3

PR #991, opened 2022-01-13
Fix inconsistent openssl dependency in cmake manifest

alchen:
Fixes https://github.com/facebook/watchman/issues/990

All manifests, except `cmake`, specifies `[dependencies.not(os=linux)]` for `openssl`. The newly added dependency otherwise causes a "project openssl has no builder" error when building on Linux distros.


chadaustin:
Thanks. I think this was also fixed by https://github.com/facebook/watchman/commit/7c31dfdb1c8c44d261447e9b03b79492c025a9e3, but I'll cc @ahornby

alchen:
@chadaustin Yes, removing the dependency altogether is as good as fixing it for Linux. Feel free to close this as needed.

genevievehelsel:
closing per comments

PR #1004, opened 2022-02-28
pywatchman: Print stderr if there was an error

jez:
I was running watchman-make at the command line and getting a backtrace that said "command failed."

It turns out that somehow my terminal process had a `nice` value of 5, and was failing the check that `watchman` has for `nice_value > min_acceptable_nice_value`, but I couldn't see that message because it was getting swallowed.

Seems that if we know we're about to make the process crash by raising an exception, we may as well dump a little extra debugging information on our way out.

### DIFF
--- a/watchman/python/pywatchman/__init__.py
+++ b/watchman/python/pywatchman/__init__.py
@@ -973,6 +973,8 @@ class client(object):
         exitcode = p.poll()
 
         if exitcode:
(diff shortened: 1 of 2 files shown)


jez:
I still would appreciate the contents of this change merged but I want a small list of open pull requests more, so I'm closing this. Feel free to re-open if you'd like to action it.

PR #1005, opened 2022-02-28
Show nice_value and min_acceptable_nice_value

jez:
This would have been helpful when debugging something recently, where I was
running watchman from a script inside a script, where it wasn't easy to stick a
call to `nice` to check the current value.


jez:
I still would appreciate the contents of this change merged but I want a small list of open pull requests more, so I'm closing this. Feel free to re-open if you'd like to action it.

chadaustin:
Apologies for our lack of response. Reimporting as #1109 1109

PR #1025, opened 2022-05-24
Fix return type in `get_linux_type()`

whippiii:
This function is expected to return a tuple of 3 elements. This doesn't happen when it fails opening the `/etc/os-release` file.


chadaustin:
Thanks! What OS did you repro this on?

whippiii:
@chadaustin Sorry for the late reply. I could reproduce it on Void Linux. While trying to build `watchman` locally using its [source packages collection](https://github.com/void-linux/void-packages/).

PR #1034, opened 2022-06-05
Fix synthesize arg in pywatchman.client.capabilityCheck

herg:
`pywatchman.capabilities.synthesize` expects an `opts` dict of the shape `{"required": list[str], "optional": list[str]}`.

`pywatchman.client.capabilityCheck` is incorrectly passing an optional `list[str]` as the `opts` arg,

### DIFF
--- a/watchman/python/pywatchman/__init__.py
+++ b/watchman/python/pywatchman/__init__.py
@@ -1164,14 +1164,13 @@ class client(object):
 
     def capabilityCheck(self, optional=None, required=None):
         """Perform a server capability check"""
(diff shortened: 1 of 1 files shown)


orf:
@chadaustin this is currently breaking all Django usages of watchman (https://github.com/django/django/blob/main/django/utils/autoreload.py#L628) - is it possible to get this tiny fix looked at?

orf:
Thank you!

SebCorbin:
Could we have a release of pywatchman to get this fix live?

PR #1036, opened 2022-06-16
watchman: docs: document Git SCM support

sluongng:
Git SCM support has been added since 2021.08.30 release via a pull
request by @arxanas. (1)

Update website documentation to reflect that Git support has been added
additionally to Mercurial.

(1): https://github.com/facebook/watchman/pull/934

### DIFF
--- a/website/_docs/scm-query.md
+++ b/website/_docs/scm-query.md
@@ -7,14 +7,18 @@ permalink: docs/scm-query.html
 redirect_from: docs/scm-query/
 ---
 
+*Since 2021.08.30*
+
+[Git support](https://github.com/facebook/watchman/pull/934) has been added 🎊
+
 *Since 4.9.*
 
 *The [capability](/watchman/docs/capabilities.html) name associated with this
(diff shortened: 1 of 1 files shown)


arxanas:
FYI the Git support is buggy ☹️. IIRC, there's some command in the API that needs to get the changes since a given commit, but my implementation only gets the *committed* changes, not the uncommitted changes.

sluongng:
> FYI the Git support is buggy ☹️. IIRC, there's some command in the API that needs to get the changes since a given commit, but my implementation only gets the _committed_ changes, not the uncommitted changes.

Getting the doc out should help us increase the contribution toward this?

Do you want to open a separate issue for tracking the bug?

sluongng:
Not sure why but this was not reflected in https://facebook.github.io/watchman/docs/scm-query.html

@chadaustin do you have any idea?

chadaustin:
Ah, I forgot to deploy the website. Should be out now. :)

PR #1042, opened 2022-07-20
fix(pywatchman): remove unreachable statement

ElijahPepe:
In the _parseEncoding() function, if the encoding is set to "bser-v1", only a WatchmanError is raised, not the return statements below it, which are unreachable due to the hierarchy of return and raise in Python. The inclusion of these return statements seems like a mistake.


chadaustin:
Thanks! I assume this is a Python 2 -> 3 migration wart.

PR #1052, opened 2022-08-24
buckconfig: remove all Watchman exceptions

chadaustin:
Summary:
A few of the exceptions for Watchman inn the buckconfig were out of date, but
some were still legit. Since there is an effort to remove all these exceptions,
let's fix the ones that were legit so we can remove all the exceptions.

Reviewed By: fanzeyi

Differential Revision: D38984208


genevievehelsel:
This has landed - according to internal signal [D38984208](https://www.internalfb.com/diff/D38984208)

PR #1087, opened 2023-01-20
rust example: remove unused Result

qm3ster:
main will never return an Err


fanzeyi:
@qm3ster thanks for contributing. Please sign the SLA above so I can accept your PR. Thank you.

qm3ster:
@fanzeyi it is done

PR #1103, opened 2023-02-23
🐛 Fix flaky test_sinceRelativeRoot

### LINKED ISSUE #1100: `test_sinceRelativeRoot` seems flaky
I'm trying to build from source, and I keep getting flaky results on https://github.com/facebook/watchman/blob/0b238b069f54a4de27c35a3e2352c37bccc12d60/watchman/integration/test_since.py#L90

While debugging, I added some log lines, and it immediately stopped flaking ([Heisenbug](https://en.wikipedia.org/wiki/Heisenbug) flag!)

I haven't fully looked into the implementation to understand this 100%, but could it be that:

https://github.com/facebook/watchman/blob/0b238b069f54a4de27c35a3e2352c37bccc12d60/watchman/integration/test_since.py#L78-L89

can flake if not enough time passes between the assignment to `clock` and the `query`?

I've locally patched this to mock the `clock` with an older time, and it stopped flaking

marianosimone:
Adjust the `clock` before querying, as `mkdir` and `touch` sometimes are so fast that make the `query` have the same time as `clock`, and not see the changes.

Fixes #1100

### DIFF
--- a/watchman/integration/test_since.py
+++ b/watchman/integration/test_since.py
@@ -82,10 +82,14 @@ class TestSince(WatchmanTestCase.WatchmanTestCase):
         self.touchRelative(os.path.join(root, "subdir"), "foo")
         self.assertFileList(root, files=["a", "subdir", "subdir/foo"])
 
+        _c, start_time, pid, root_number, ticks = clock.split(":")
+        # Move the clock back to avoid flakiness when mkdir/touch are fast enough to make the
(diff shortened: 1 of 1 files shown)


fanzeyi:
Thanks! This looks reasonable. I'll get it merge tomorrow at work.

We really should have a fake test clock mechanism in Watchman.

fanzeyi (approved):

PR #1144, opened 2023-05-26
Update index.js

oygen87:
fix typo


genevievehelsel:
Please complete the Facebook CLA Check so we can continue with importing and applying this change :)

PR #1156, opened 2023-08-09
Update install.md

Fares-Fuad:
Add the simple way for install via apt


chadaustin:
As described in the sentence following this patch, the Watchman package supplied by Ubuntu is old and insecure. We don't want to encourage it.

PR #1183, opened 2024-01-09
Use achives.boost.io instead of jfrog for fetching boost

ASchneidman:
Summary: AIRStore's contbuild broke a few days ago with a failure to unpack the boost tarball. This is because boost's jfrog subscription expired. The recommended workaround is to use `archive.boost.io` instead. The checksums are the same. https://github.com/facebook/react-native/issues/42180

Reviewed By: chadaustin

Differential Revision: D52612598


ahornby:
done

PR #1215, opened 2024-04-22
WIP - Implement `maximum_results` option

yamadapc:
When working on large mono-repos some queries might take a very long time due to listing all files on the project.

This waiting time is a combination of building the results array and serialising it back.

This commit adds a query option `maximum_results`, which allows clients to ask the server to truncate results above a certain threshold number.

On that case, a flag is sent back indicating to the client results have been truncated.

### DIFF
--- a/watchman/InMemoryView.cpp
+++ b/watchman/InMemoryView.cpp
@@ -588,6 +588,10 @@ void InMemoryView::timeGenerator(const Query* query, QueryContext* ctx) const {
       continue;
     }
 
+    if (ctx->num_results_over_maximum > 0) {
+      continue;
+    }
+
     w_query_process_file(
         query, ctx, std::make_unique<InMemoryFileResult>(f, caches_));
   }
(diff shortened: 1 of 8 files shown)


yamadapc:
Hello, could I get an opinion on this diff?

yamadapc:
Hey :wave:

genevievehelsel:
> Hey 👋

followed up in the Sapling discord (nice idea asking there :) ) but copy-pasting here for wider visibility:

hi from the watchman team - I think this looks like a good idea 🙂 we've got a similar flag like @quark mentioned for a watcher we use often (watcher/eden.cpp) for the same reasons you mentioned in your summary

on the making the eden_file_count_threshold_for_fresh_instannce more universal, the code pointer for that is here: 

https://github.com/facebook/watchman/blob/main/watchman/watcher/eden.cpp#L760

It would be a lot cleaner IMO if we could just leverage the existing fresh_instance code for this, in a similar manner to how its done for the eden watcher, especially since some clients might already know how to handle fresh instances and they'd have to relearn how to handle a new flag if they wanted to use this feature

we have an integration test for this flag for the eden watcher here: https://github.com/facebook/watchman/blob/main/watchman/integration/eden/test_eden_since.py#L296

if you went the route of returning fresh instances if threshold is reached, you could just make a similar test to the above but for a non-eden watcher 🙂

PR #1221, opened 2024-06-11
feat: add rust client apis for trigger, trigger-list, trigger-del

fowles:


fowles:
I added trigger-list as well.  I decided to punt on deserialize the more complicated portions of the request.

fowles:
friendly ping, I would love to get this in so I can use it in https://github.com/martinvonz/jj/issues/3855

watchman/rust/watchman_client/src/pdu.rs
```diff
@@ -559,6 +560,121 @@ pub struct SubscribeResponse {
...
+    pub expression: Option<Expr>,
+
+    /// Configure the way `stdin` is configured for the executed trigger.
+    #[serde(
+        default,
+        skip_serializing_if = "TriggerStdinConfig::is_devnull",
+        serialize_with = "TriggerStdinConfig::serialize",
+        skip_deserializing
+    )]
+    pub stdin: TriggerStdinConfig,
```
genevievehelsel, line 591:
```suggestion
    pub stdin: Option<TriggerStdinConfig>,
```

nit, it looks like you have it set up correctly to default to `/dev/null`, but Watchman knows how to default itself to `/dev/null` if this is omitted, so it probably is best to let Watchman handle this being omitted rather than defaulting it in the client api (in case Watchman changes the behavior of this being omitted): 

https://github.com/facebook/watchman/blob/f6ee0299b7c6dc15ce694d2ead92fae9bce96a22/watchman/TriggerCommand.cpp#L349-L352

.gitignore
```diff
@@ -81,8 +81,8 @@ libtool
 ltmain.sh
 cppclient/watchmanclient.pc
 /website/.jekyll-metadata
-/rust/**/Cargo.lock
-/rust/**/target/
+/**/rust/**/Cargo.lock
+/**/rust/**/target/
```
genevievehelsel, line 85: (copying for @jdelliot from Phabricator)

Can we just use this:
```suggestion
rust/**/Cargo.lock
rust/**/target/
```

watchman/rust/watchman_client/src/pdu.rs
```diff
@@ -559,6 +560,121 @@ pub struct SubscribeResponse {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_files_stdin: Option<u64>,
+
+    /// Specifies the working directory that will be set prior to spawning the
+    /// process. The default is to set the working directory to the watched
+    /// root. The value of this property is a string that will be interpreted
+    /// relative to the watched root.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub chdir: Option<String>,
+}
```
genevievehelsel, line 614: Thoughts on supporting `relative_root`?

```suggestion

    /// Since 3.4. Evaluate the trigger with respect to a path within a watched root
    #[serde(skip_serializing_if = "Option::is_none")]
    pub relative_root: Option<String>,
}
```

genevievehelsel (changes requested):
Thanks for the PR - overall lgtm, left a few inlines, lmk what you think :)

watchman/rust/watchman_client/src/pdu.rs
fowles: The use of `skip_serializing_if = "TriggerStdinConfig::is_devnull",` on line 587 accomplishes the same thing but means that I don't have to deal with an extra layer of wrap/unwrap in all the surrounding code

.gitignore
fowles: That does not work for me.  I see all the build artifacts included when that happens.

watchman/rust/watchman_client/src/pdu.rs
fowles: added

watchman/rust/watchman_client/src/pdu.rs
fowles: oh, I understand what you mean.  Will change it.

.gitignore
quark-zju: You can add a `.gitignore` inside `rust/`, so the rules can be `Cargo.lock` and `target` there, which is cheaper to handle.

fowles: that is awesome!  I had no idea you could used nested .gitignores.  I have put it in the `rust` subdirectory instead of putting one in rust/watchman_client and one in rust/serde

quark-zju:
Somehow the internal CI performs clippy checks, which this PR did not pass. Could you try to rebase past https://github.com/facebook/watchman/commit/861531ff9c6318324c64f2bf5a1dae266280e106 to address the clippy issues?

fowles:
> Somehow the internal CI performs clippy checks, which this PR did not pass. Could you try to rebase past [861531f](https://github.com/facebook/watchman/commit/861531ff9c6318324c64f2bf5a1dae266280e106) to address the clippy issues?

Done!

PR #1229, opened 2024-06-26
fix: update watchman_client version

chenrui333:
Seeing some build error as below:

```
error: failed to select a version for the requirement `watchman_client = "^0.8.0"`
candidate versions found which didn't match: 0.9.0
```

relates to https://github.com/Homebrew/homebrew-core/pull/175551


genevievehelsel:
This is an autogenerated file so internally CI is not happy with the manual change, I sent an internal PR to fix the source of truth and regenerate this file

genevievehelsel:
see: https://github.com/facebook/watchman/commit/aee833d955ae1418eb592778c80231b9f5734f1b

chenrui333:
Works for me, thanks!

PR #1252, opened 2024-11-20
Don't enable `future` crate's `compat` feature

martinvonz:


martinvonz:
I checked with some open-source team at Google about the process for signing the CLA It sounds too complicated to be worth it. So I'll just close this PR :(

Issue #1254, opened 2024-11-26
failed to build watchman 2024.11.25.00


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>error build log</summary>

```
     Compiling sysinfo v0.30.13
  error[E0599]: no method named `len` found for type `i8` in the current scope
     --> /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-6f17d22bba15001f/sysinfo-0.30.13/src/unix/apple/macos/process.rs:436:53
      |
  436 |         Some(node.vip_path.len() * node.vip_path[0].len()),
      |                                                     ^^^
      |
  help: there is a method `le` with a similar name, but with different arguments
     --> /private/tmp/rust-20241019-8559-8pur06/rustc-1.82.0-src/library/core/src/cmp.rs:1196:5
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/12016510835/job/33509504050
relates to Homebrew/homebrew-core#198962

chenrui333:
issue is resolved by now

PR #1255, opened 2024-11-27
fix: bump sysinfo to 0.32.1

chenrui333:
seeing some build failure as in https://github.com/GuillaumeGomez/sysinfo/issues/1392 while building the latest release.

Bumping sysinfo crate to 0.32.1 fixed the issue, relates to https://github.com/Homebrew/homebrew-core/pull/198962

fixes https://github.com/facebook/watchman/issues/1254


chenrui333:
any update on the thread? thanks!

chenrui333:
@AjanthanAsogamoorthy can you also help with this PR? thanks!

ahornby:
@chenrui333 is this because you are picking up a newer libc crate or some other updated dep, and do you have a workaround to lock libc or whatever dep is causing issues in the homebrew builds?

chenrui333:
I think with the newer libc crate release, I dont need the patch anymore, thanks for weighing in here.

PR #1322, opened 2025-12-30
Use /proc/self instead of /proc/<pid> in getOpenedPath

ptarjan:
When running in a PID namespace (common in containers), watchman fails
with "need /proc to be mounted!" because it accesses /proc/<PID>/fd/X
using the namespace PID, but /proc may be mounted from the host with
different PID numbering.

The fix is to use /proc/self/fd/X instead of /proc/<getpid()>/fd/X
when resolving file descriptor paths. The /proc/self symlink is always
resolved by the kernel to the current process's /proc entry, regardless
of PID namespace boundaries.

This issue manifests when:
- Process is in a PID namespace (e.g., Docker container)
- /proc is mounted from host or a different namespace
- watchman tries to resolve paths via /proc/<PID>/fd

Error seen:
  getOpenedPath: need /proc to be mounted!: Function not implemented

Affected platforms: Linux containers with PID namespace isolation

### DIFF
--- a/watchman/fs/FileDescriptor.cpp
+++ b/watchman/fs/FileDescriptor.cpp
@@ -7,7 +7,6 @@
 
 #include "watchman/fs/FileDescriptor.h"
 #include <folly/String.h>
-#include <folly/system/Pid.h>
 #include <system_error>
 #include "watchman/fs/FSDetect.h"
 #include "watchman/fs/FileInformation.h"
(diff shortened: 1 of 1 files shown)


ptarjan:
> @ptarjan has updated the pull request. You must [reimport the pull request](https://www.internalfb.com/intern/opensource/github/pull_request/confirm/import/865772219744942/) before landing.

I'm doing this to re-run the flaky test

ptarjan:
I'm giving up on CI. It is failing on `main`. The code compiles and works on my machine and I tested in my fleet and it works great.

ptarjan:
@genevievehelsel is there anything else you need from me for this?

